PR #11, opened 2017-05-31
Add initial spec based on explainer

beaufortfrancois:
@mounirlamouri As promised, here it is.

FYI, bikeshed is not happy:

```
LINK ERROR: No 'dfn' refs found for 'htmlvideoelement'.
<a data-link-type="dfn" data-lt="HTMLVideoElement"><code>HTMLVideoElement</code></a>
LINK ERROR: No 'dfn' refs found for 'remote playback specification'.
<a data-link-type="dfn" data-lt="Remote Playback specification">Remote Playback specification</a>
LINK ERROR: No 'dfn' refs found for 'feature policy specification'.
<a data-link-type="dfn" data-lt="Feature Policy specification">Feature Policy specification</a>
```

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,0 +1,265 @@
+<pre class='metadata'>
+Title: Picture In Picture
+Shortname: picture-in-picture
+Level: 1
+Status: ED
+ED: https://wicg.github.io/picture-in-picture
+Group: WICG
+Editor: François Beaufort, Google, fbeaufort@google.com
+Editor: Mounir Lamouri, Google, mlamouri@google.com
+Abstract: This specification intends to provide APIs to allow websites to
+Abstract: create a floating video window over the desktop.
+</pre>
+
+# Introduction # {#intro}
+
+<em>This section is non-normative.</em>
+
+Many users want to continue consuming media while they interact with other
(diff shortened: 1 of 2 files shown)


mounirlamouri:
Maybe it shouldn't be a literal copy-paste of the explainer :)

The errors you see are because bikeshed tries to linkify thinks that are inside anchor tags like `<a>foo</a>`. It links to whatever is defined as `foo` with `<dfn>foo</a>`.

What about following a template using Media Session, Remote Playback, or Media Capabilities and fill in the blank with the Picture in Picture explainer? :)

beaufortfrancois:
@mounirlamouri I'm still having some trouble with 🚲 but I think it is in a better shape. WDYT?

mounirlamouri:
In general, I would recommend against using the markdown syntax in bikeshed. I find it confusing to read markdown and HTML mixed together. Though, it might be because of my previous life as spec editor working with respec.

Is there something in particular that causes you trouble and you want me to look at?

beaufortfrancois:
I tend to use MD as much as possible as I find it much easier to read and less prone to HTML errors. I'll use HTML only when I have to (descriptions lists, IDL, etc.)

PR #24, opened 2017-06-09
Emphasize the fact the window is always on top

beaufortfrancois:
R: @mounirlamouri 
FIX: https://github.com/WICG/picture-in-picture/issues/23


mounirlamouri:
lgtm, thanks!

PR #27, opened 2017-12-15
Detail the reasons why PIP can't be triggered

beaufortfrancois:
FIX https://github.com/WICG/picture-in-picture/issues/22

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -129,13 +129,11 @@ and |options| is invoked, the user agent MUST run the following steps:
     top</dfn>. It consistently stays above most other windows.
 4. Let |pipVideo| be the video contained in the |pipWindow|.
 5. If a |pipWindow| is not already created, create one.
-6. If one can't be created because user agent blocked feature for instance,
-    throw a {{NotAllowedError}} and abort these steps.
-7. Apply |options| {{width}}, {{height}}, {{PictureInPictureOptions/top}}, and
-    {{start}} to the |pipWindow| dimensions and position. If some |options|
-    can't be applied, it SHOULD apply latest ones that were successfully set if
-    any.
-8. Render |video| frames in the |pipVideo|.
+6. OPTIONALLY, apply |options| {{width}}, {{height}},
+    {{PictureInPictureOptions/top}}, and {{start}} to the |pipWindow| dimensions
+    and position. If some |options| can't be applied, it applies latest ones that
+    were successfully set if any.
+7. Render |video| frames in the |pipVideo|.
 
 It is RECOMMENDED that the |video| frames are not rendered in the page and in
 the |pipVideo| at the same time but if they are, they MUST be kept in sync.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -122,16 +122,20 @@ and |options| is invoked, the user agent MUST run the following steps:
...
+    steps.
+3. Let |pipWindow| be the Picture In Picture window that is <dfn>always on
     top</dfn>. It consistently stays above most other windows.
-3. Let |pipVideo| be the video contained in the |pipWindow|.
-4. If a |pipWindow| is not already created, create one.
-5. If one can't be created, throw a {{UnknownError}} and abort these steps.
-6. Apply |options| {{width}}, {{height}}, {{PictureInPictureOptions/top}}, and
+4. Let |pipVideo| be the video contained in the |pipWindow|.
+5. If a |pipWindow| is not already created, create one.
+6. If one can't be created because user agent blocked feature for instance,
```
mounirlamouri: What do you mean by "because of user agent blocked feature"?

index.bs
```diff
@@ -122,16 +122,20 @@ and |options| is invoked, the user agent MUST run the following steps:
...
+    throw a {{NotAllowedError}} and abort these steps.
+7. Apply |options| {{width}}, {{height}}, {{PictureInPictureOptions/top}}, and
```
mounirlamouri: Maybe add "OPTIONALLY,", here?

index.bs
```diff
@@ -122,16 +122,20 @@ and |options| is invoked, the user agent MUST run the following steps:
...
     {{start}} to the |pipWindow| dimensions and position. If some |options|
     can't be applied, it SHOULD apply latest ones that were successfully set if
```
mounirlamouri: If we add "Optionnaly", maybe the SHOULD could be dropped?

index.bs

index.bs

index.bs
beaufortfrancois: I've removed this step eventually, as we already check for `pictureInPictureEnabled` and return a promise rejected with `NotSupportedError`. WDYT?

beaufortfrancois:
@mounirlamouri I've addressed your feedback above.

index.bs
```diff
@@ -122,15 +122,17 @@ and |options| is invoked, the user agent MUST run the following steps:
...
-5. If one can't be created, throw a {{UnknownError}} and abort these steps.
-6. Apply |options| {{width}}, {{height}}, {{PictureInPictureOptions/top}}, and
-    {{start}} to the |pipWindow| dimensions and position. If some |options|
-    can't be applied, it SHOULD apply latest ones that were successfully set if
-    any.
+4. Let |pipVideo| be the video contained in the |pipWindow|.
+5. If a |pipWindow| is not already created, create one.
+6. OPTIONALLY, apply |options| {{width}}, {{height}},
+    {{PictureInPictureOptions/top}}, and {{start}} to the |pipWindow| dimensions
+    and position. If some |options| can't be applied, it applies latest ones that
```
mounirlamouri: s/it applies/applies/ ?

beaufortfrancois:
@mounirlamouri Fixed.

mounirlamouri:
:+1: thanks :)

PR #28, opened 2017-12-15
Add a way to disable PiP

beaufortfrancois:
This PR is about adding a hint to disable Picture In Picture for a specific video element as some may not want this for some videos.

WDYT?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -175,7 +175,7 @@ the user agent SHOULD run these steps:
 
 1. Reject any pending promises returned by the {{requestPictureInPicture()}}
     method with {{InvalidStateError}}.
-2. If |video| is playing in Picture in Picture:
+2. If |video| is |pipVideo|:
     1. Run the <a>exit picture in picture algorithm</a>.
     2. Unset {{pictureInPictureElement}}.
     3. <a>Queue a task</a> to <a>fire a simple event</a> with the name
@@ -225,7 +225,7 @@ partial interface HTMLVideoElement {
     attribute EventHandler onleavepictureinpicture;
 
     [CEReactions]
-    attribute Boolean disablePictureInPicture;
+    attribute boolean disablePictureInPicture;
 };
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -76,8 +76,8 @@ extensible.
 <button id="pipButton"></button>
 
 <script>
-  // Hide button if Picture In Picture is not supported.
-  pipButton.hidden = !document.pictureInPictureEnabled;
+  // Hide button if Picture In Picture is not supported or disabled.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -155,6 +155,33 @@ the user agent MUST run the following steps:
...
+If the {{disablePictureInPicture}} attribute is present on the video element,
+the user agent SHOULD NOT play the video element in picture in picture or
+present any UI to do so.
+
+When the {{disablePictureInPicture}} attribute is added to a |video| element,
+the user agent SHOULD run these steps:
+
+1. Reject any pending promises returned by the {{requestPictureInPicture()}}
+    method with {{InvalidStateError}}.
+2. If |video| is playing in Picture in Picture:
```
mounirlamouri: We should be able to use more normative language here. We do expose the current PIP element, right?

jernoble:
@mounirlamouri 
> My main concern is that blocking the API with an attribute that the website would set is a bit weird.

Would it be less weird to you if the attribute reflected to a property? (Or are we not doing attribute reflection any longer?)

mounirlamouri:
I think @beaufortfrancois's proposal had a property reflecting the attribute.

In general, I'm not strongly against this and if you would ship an attribute that would fully bock the API and the UA features (UI buttons, auto-pip, etc.), I would be fine having it in Blink and therefore in the spec.

index.bs
```diff
@@ -155,6 +155,33 @@ the user agent MUST run the following steps:
...
+MUST reflect the content attribute of the same name.
+
+If the {{disablePictureInPicture}} attribute is present on the video element,
+the user agent SHOULD NOT play the video element in picture in picture or
+present any UI to do so.
+
+When the {{disablePictureInPicture}} attribute is added to a |video| element,
+the user agent SHOULD run these steps:
+
+1. Reject any pending promises returned by the {{requestPictureInPicture()}}
```
mounirlamouri: Shouldn't you add steps in the enter picture in picture algorithm to reject if the attribute is set too?

beaufortfrancois: Fixed in https://github.com/WICG/picture-in-picture/pull/28/commits/04ce8d3084193ab3c7ff7d23808eee0c59450cdb

index.bs
```diff
@@ -125,11 +125,13 @@ invoked, the user agent MUST run the following steps:
     steps.
 2. If the algorithm is not <a>triggered by user activation</a>, throw a
     {{NotAllowedError}} and abort these steps.
-3. Let |pipWindow| be the Picture In Picture window that is <dfn>always on
+3. OPTIONALLY, if the {{disablePictureInPicture}} attribute is present on
+    |video|, throw a {{NotAllowedError}} and abort these steps.
```
mounirlamouri: Shouldn't we stay consistent and use `InvalidStateError`?

PR #30, opened 2018-01-19
Trigger NotAllowedError for User Activation

beaufortfrancois:
@mounirlamouri WDYT?


mounirlamouri:
Shouldn't this be the opposite? `NotAllowerError` should be user for user activation to be consistent with autoplay and `SecurityError` for feature policy?

PR #36, opened 2018-01-23
Add Travis CI status button to README

beaufortfrancois:


mounirlamouri:
It's a bit funny to do this for a spec with Markdown but why not :)

PR #39, opened 2018-01-28
s/Element/HTMLVideoElement for pictureInPictureElement

beaufortfrancois:


mounirlamouri:
I'm not fully against this but it was on purpose to make sure we were forward compatible. I think that your change wouldn't break backward compatibility though so maybe no big deal but I'm not sure it's needed either way. With a section is the spec explaining why it's `Element` and not `HTMLVideoElement` would your concerns be addressed?

beaufortfrancois:
I'd like as much as possible our spec to reflect the intent of this first version which is that it applies only to a `HTMLVideoElement`. A section explaining why may help, but as this change wouldn't break backward compatibility, I'd suggest we simply use `HTMLVideoElement` over `Element`:

```diff
partial interface DocumentOrShadowRoot {
-  readonly attribute Element? pictureInPictureElement;
+  readonly attribute HTMLVideoElement? pictureInPictureElement;
};

PR #45, opened 2018-01-30
Explained why only HTMLVideoElement in security

beaufortfrancois:
FIX https://github.com/WICG/picture-in-picture/issues/42

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -334,8 +334,9 @@ dimensions about the current Picture-in-Picture window.
 
 <em>This section is non-normative.</em>
 
-The API only applies on {{HTMLVideoElement}} to punt security issues around
-PIP-ing arbitrary HTML content and handling events.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -332,6 +332,11 @@ dimensions about the current Picture-in-Picture window.
 
 # Security considerations # {#security-considerations}
 
+<em>This section is non-normative.</em>
+
+The API only applies on {{HTMLVideoElement}} to punt security issues around
+PIP-ing arbitrary HTML content and handling events.
```
mounirlamouri: The API applies only to {{HTMLVideoElement}} in order to start on a minimal viable product that has limited security issues. Later versions of this specification may allow PIP-ing arbitrary HTML content.

beaufortfrancois:
@mounirlamouri Fixed.

PR #48, opened 2018-01-30
Exit PiP throws InvalidStateError if no pip video

beaufortfrancois:
WDYT?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -175,15 +175,17 @@ state that is considered not compatible with Picture-in-Picture.
 When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked,
 the user agent MUST run the following steps:
 
-1. Let |pipWindow| be the current Picture-in-Picture window.
-2. Let |pipVideo| be the video contained in the |pipWindow|.
-3. Let |video| be the source video of |pipVideo| frames.
-4. Close |pipWindow|.
-5. If |pipWindow| can't be closed, throw a {{UnknownError}} and abort these
+1. If {{pictureInPictureElement}} is null, throw a {{InvalidStateError}} and abort
+    these steps.
+2. Let |pipWindow| be the current Picture-in-Picture window.
+3. Let |pipVideo| be the video contained in the |pipWindow|.
+4. Let |video| be the source video of |pipVideo| frames.
+5. Close |pipWindow|.
+6. If |pipWindow| can't be closed, throw a {{UnknownError}} and abort these
     steps.
-6. Render |pipVideo| frames in the |video|.
-7. Unset {{pictureInPictureElement}}.
-8. <a>Queue a task</a> to <a>fire a simple event</a> with the name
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -175,15 +175,17 @@ state that is considered not compatible with Picture-in-Picture.
 When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked,
 the user agent MUST run the following steps:
 
-1. Let |pipWindow| be the current Picture-in-Picture window.
-2. Let |pipVideo| be the video contained in the |pipWindow|.
-3. Let |video| be the source video of |pipVideo| frames.
-4. Close |pipWindow|.
-5. If |pipWindow| can't be closed, throw a {{UnknownError}} and abort these
+1. If {{pictureInPictureElement}} is null, throw a {{NotFoundError}} and abort
```
mounirlamouri: WDYT of `InvalidStateError`?

index.bs
```diff
@@ -175,15 +175,17 @@ state that is considered not compatible with Picture-in-Picture.
...
+    these steps.
+2. Let |pipWindow| be the current Picture-in-Picture window.
+3. Let |pipVideo| be the video contained in the |pipWindow|.
+4. Let |video| be the source video of |pipVideo| frames.
+5. Close |pipWindow|.
+6. If |pipWindow| can't be closed, throw a {{UnknownError}} and abort these
```
mounirlamouri: Is it something we know can happen?

index.bs
beaufortfrancois: LGTM

index.bs
beaufortfrancois: I guess everything can happen but I don't have anything in mind that can cause a |pipWindow| to not close.

mounirlamouri: I think we should drop this. Anything "unexpected" would be handled by the implementations. If we don't have real cases justifying this to happen, we shouldn't worry about it.

beaufortfrancois: Noted.

beaufortfrancois:
@mounirlamouri Done.

index.bs
```diff
@@ -175,12 +175,12 @@ state that is considered not compatible with Picture-in-Picture.
...
-2. Let |pipVideo| be the video contained in the |pipWindow|.
-3. Let |video| be the source video of |pipVideo| frames.
-4. Close |pipWindow|.
-5. If |pipWindow| can't be closed, throw a {{UnknownError}} and abort these
-    steps.
+1. If {{pictureInPictureElement}} is null, throw a {{InvalidStateError}} and abort
+    these steps.
+2. Let |pipWindow| be the current Picture-in-Picture window.
+3. Let |pipVideo| be the video contained in the |pipWindow|.
+4. Let |video| be the source video of |pipVideo| frames.
```
mounirlamouri: This difference between |pipVideo|, |video| and |pictureInPictureElement| is weak. Probably orthogonal to this change though.

PR #49, opened 2018-01-31
Reorder error steps in request PiP

beaufortfrancois:
I'm reordering error steps so that web developer can take much easier some actions on this.

1. PiP is not enabled (you can't do anything)
2. Feature Policy prevents you to use PiP (update it if you can)
3. Video attribute prevents you to use PiP (remove it)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -142,8 +142,8 @@ extensible.
 When the <dfn>request Picture-in-Picture algorithm</dfn> with |video| is
 invoked, the user agent MUST run the following steps:
 
-1. If {{pictureInPictureEnabled}} is |false|, throw a {{NotSupportedError}} and
-    abort these steps.
+1. If <a>Picture-in-Picture support</a> is <code>false</code>, throw a
+    {{NotSupportedError}} and abort these steps.
 2. If document is not allowed to use the <a>policy-controlled feature</a> named
     <code>"picture-in-picture"</code>, throw a {{SecurityError}} and abort these
     steps.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -277,9 +277,9 @@ partial interface Document {
 </pre>
 
 The {{pictureInPictureEnabled}} attribute's getter must return
-<code>true</code> if the <a>context object</a> is <a>allowed to use</a> the
-feature indicated by attribute name <code>picture-in-picture</code> and
-<a>Picture-in-Picture is supported</a>, and false otherwise.
+<code>true</code> if <a>Picture-in-Picture is supported</a> and the <a>context
+object</a> is <a>allowed to use</a> the feature indicated by attribute name
+<code>picture-in-picture</code>, and false otherwise.
```
mounirlamouri: Isn't this about the feature policy? Does that mean that the steps above (checking for `pictureInPictureEnabled` would check for the feature policy too implicitly, making step 2 moot?

beaufortfrancois: You're absolutely right! Good catch.

beaufortfrancois:
@mounirlamouri Feedback addressed. WDYT?

index.bs
```diff
@@ -277,12 +277,14 @@ partial interface Document {
...
-
-<dfn>Picture-in-Picture is supported</dfn> if there is no
-previously-established user preference, security risk, or platform limitation.
+<code>true</code> if <a>Picture-in-Picture support</a> is <code>true</code> and
+the <a>context object</a> is <a>allowed to use</a> the feature indicated by
+attribute name <code>picture-in-picture</code>, and <code>false</code>
+otherwise.
+
+<dfn>Picture-in-Picture support</dfn> is <code>true</code> if there is no
+previously-established user preference, security risk, or platform limitation,
```
mounirlamouri: security risk? That's unclear. Restrictions? Though maybe keep to "platform limitation"?

PR #53, opened 2018-02-02
Use unsigned long for pip Window dimensions

beaufortfrancois:
According to https://www.w3.org/TR/WebIDL-1/#idl-long,

> The long type is a signed integer type that has values in the range [−2147483648, 2147483647].
> The unsigned long type is an unsigned integer type that has values in the range [0, 4294967295].

I believe we should use "unsigned long" then for pipWindow width and height.


beaufortfrancois:
Noted.

PR #55, opened 2018-02-05
s/HTMLVideoElement/Element for pictureInPictureElement

beaufortfrancois:
It turns out `Element` is required for `pictureInPictureElement` as it needs to return shadow root elements as well.

This CL updates spec and explainer.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -307,11 +307,12 @@ partial interface DocumentOrShadowRoot {
 The {{pictureInPictureElement}} attribute's getter must run these steps:
 
 1. If the <a>context object</a> is a <a for=/>shadow root</a> and its
-    <a for=DocumentFragment>host</a> is not <a>connected</a>, then return null.
+    <a for=DocumentFragment>host</a> is not <a>connected</a>, return null and
+    abort these steps.
 2. Let |candidate| be the result of <a>retargeting</a> Picture-in-Picture
     element against the <a>context object</a>.
 3. If |candidate| and the <a>context object</a> are in the same <a>tree</a>,
-    then return |candidate|.
+    return |candidate| and abort these steps.
 4. Return null.
 
 ## <code>PictureInPictureWindow</code> ## {#picture-in-picture-window}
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -300,15 +300,19 @@ parallel</a>:
...
+  readonly attribute Element? pictureInPictureElement;
 };
 </pre>
 
 The {{pictureInPictureElement}} attribute's getter must run these steps:
 
 1. If the <a>context object</a> is a <a for=/>shadow root</a> and its
     <a for=DocumentFragment>host</a> is not <a>connected</a>, then return null.
-2. Return the Picture-in-Picture video element.
+2. Let |candidate| be the result of <a>retargeting</a> Picture-in-Picture
```
mounirlamouri: Is "retargeting" defined in the spec?

index.bs
```diff
@@ -300,15 +300,19 @@ parallel</a>:
...
+    element against the <a>context object</a>.
+3. If |candidate| and the <a>context object</a> are in the same <a>tree</a>,
+    then return |candidate|.
```
mounirlamouri: Usually, we add "and abort these steps" but I see I forgot this comment for the other changes :(

index.bs
beaufortfrancois: It is. See https://dom.spec.whatwg.org/#retarget

index.bs

beaufortfrancois:
@mounirlamouri Your feedback is addressed. Thanks for the review!

index.bs
```diff
@@ -300,15 +300,20 @@ parallel</a>:
...
 <pre class="idl">
 partial interface DocumentOrShadowRoot {
-  readonly attribute HTMLVideoElement? pictureInPictureElement;
+  readonly attribute Element? pictureInPictureElement;
 };
 </pre>
 
 The {{pictureInPictureElement}} attribute's getter must run these steps:
 
 1. If the <a>context object</a> is a <a for=/>shadow root</a> and its
```
hayatoito: step 1's wording be redundant.

You can say simply:
- If the context object is not *connected*, return null and abort these steps.

beaufortfrancois: Done. Thank you @hayatoito!

PR #59, opened 2018-02-06
Reset Picture-in-Picture window dimensions

### LINKED ISSUE #54: What does happen to PictureInPictureWindow object when exitPictureInPicture?
What does happen to PictureInPictureWindow object when exitPictureInPicture?
Should it be set to null?
Should its `width` and `height` dimensions reset? If so, would 0 makes sense?

What do you think @mounirlamouri?

beaufortfrancois:
What do you think of this CL?

FIX #54

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/59.html" title="Last updated on Feb 12, 2018, 12:36 PM GMT (432b457)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/59/f69a02a...432b457.html" title="Last updated on Feb 12, 2018, 12:36 PM GMT (432b457)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -320,8 +320,8 @@ interface PictureInPictureWindow {
 </pre>
 
 A {{PictureInPictureWindow}} instance represents a <a>Picture-in-Picture
-window</a> associated with an {{Element}}. When instantiated, an instance of
-{{PictureInPictureWindow}} has its |state| set to |opened|.
+window</a> associated with an {{HTMLVideoElement}}. When instantiated, an
+instance of {{PictureInPictureWindow}} has its |state| set to |opened|.
 
 When the <dfn>close window algorithm</dfn> with an instance of
 {{PictureInPictureWindow}} is invoked, its |state| is set to |closed|.
@@ -330,7 +330,7 @@ The {{width}} attribute MUST return the width in <a lt=px value>CSS pixels</a> o
 <a>Picture-in-Picture window</a> associated with {{pictureInPictureElement}} if
 the |state| is |opened|. Otherwise, it MUST return 0.
 
-The {{height}} attribute is the height in <a lt=px value>CSS pixels</a> of the
+The {{height}} attribute MUST return the height in <a lt=px value>CSS pixels</a> of the
 <a>Picture-in-Picture window</a> associated with {{pictureInPictureElement}} if
 the |state| is |opened|. Otherwise, it MUST return 0.
 
--- a/index.bs
+++ b/index.bs
@@ -154,10 +154,8 @@ invoked, the user agent MUST run the following steps:
     {{NotAllowedError}} and abort these steps.
 5. Set {{pictureInPictureElement}} to |video|.
 6. Let <dfn>Picture-in-Picture window</dfn> be a new instance of
-    {{PictureInPictureWindow}} with its |state| set to |opened|.
-7. Associate {{pictureInPictureElement}} with the <a>Picture-in-Picture
-    window</a>.
-8. <a>Queue a task</a> to <a>fire an event</a> with the name
+    {{PictureInPictureWindow}} associated with {{pictureInPictureElement}}.
+7. <a>Queue a task</a> to <a>fire an event</a> with the name
     {{enterpictureinpicture}} at the |video| with its {{bubbles}} attribute
     initialized to true.
 
@@ -178,8 +176,8 @@ the user agent MUST run the following steps:
 
 1. If {{pictureInPictureElement}} is null, throw a {{InvalidStateError}} and abort
     these steps.
-2. Set the <a>Picture-in-Picture window</a> associated with
-    {{pictureInPictureElement}} |state| to |closed|.
+2. Run the <a>close window algorithm</a> with the <a>Picture-in-Picture
+    window</a> associated with {{pictureInPictureElement}}.
 3. Unset {{pictureInPictureElement}}.
 4. <a>Queue a task</a> to <a>fire an event</a> with the name
     {{leavepictureinpicture}} at the |video| with its {{bubbles}} attribute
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -321,8 +323,13 @@ interface PictureInPictureWindow {
...
 
-The {{PictureInPictureWindow}} is a simple proxy to get information such as
-dimensions about the current Picture-in-Picture window.
+The {{PictureInPictureWindow}} interface is a simple proxy to get information
+such as dimensions about the current <a>Picture-in-Picture window</a>.
+
+<dfn>Picture-in-Picture window</dfn> initial |state| is |closed|.
+
+When <a>Picture-in-Picture window</a> |state| is |closed|, its dimensions
+{{width}} and {{height}} SHOULD be set to 0.
```
mounirlamouri: Shouldn't you define how `width` and `height` should work. Basically what they should return and for both mention the `closed` state as the case where 0 should be returned?

index.bs
```diff
@@ -158,7 +158,8 @@ invoked, the user agent MUST run the following steps:
 7. If a |pipWindow| is not already created, create one.
 8. Render |video| frames in the |pipVideo|.
 9. Set {{pictureInPictureElement}} to |video|.
-10. <a>Queue a task</a> to <a>fire an event</a> with the name
+10. Set <a>Picture-in-Picture window</a> |state| to |opened|.
+11. <a>Queue a task</a> to <a>fire an event</a> with the name
```
mounirlamouri: Maybe create an instance? Do we want to handle multi-window case here?

beaufortfrancois:
@mounirlamouri I think* I've addressed your feedback. Let me know what you think.

index.bs
```diff
@@ -152,18 +152,18 @@ invoked, the user agent MUST run the following steps:
...
-6. Let |pipVideo| be the video contained in the |pipWindow|.
-7. If a |pipWindow| is not already created, create one.
-8. Render |video| frames in the |pipVideo|.
-9. Set {{pictureInPictureElement}} to |video|.
-10. <a>Queue a task</a> to <a>fire an event</a> with the name
+5. Set {{pictureInPictureElement}} to |video|.
+6. Let <dfn>Picture-in-Picture window</dfn> be a new instance of
+    {{PictureInPictureWindow}} with its |state| set to |opened|.
+7. Associate {{pictureInPictureElement}} with the <a>Picture-in-Picture
+    window</a>.
```
mounirlamouri: I guess you could merge 6 and 7 by saying "associated to {{pictureInPictureElement}}"

index.bs
```diff
@@ -152,18 +152,18 @@ invoked, the user agent MUST run the following steps:
...
-5. Let |pipWindow| be the Picture-in-Picture window that is <dfn>always on
-    top</dfn>. It consistently stays above most other windows.
-6. Let |pipVideo| be the video contained in the |pipWindow|.
-7. If a |pipWindow| is not already created, create one.
-8. Render |video| frames in the |pipVideo|.
-9. Set {{pictureInPictureElement}} to |video|.
-10. <a>Queue a task</a> to <a>fire an event</a> with the name
+5. Set {{pictureInPictureElement}} to |video|.
+6. Let <dfn>Picture-in-Picture window</dfn> be a new instance of
+    {{PictureInPictureWindow}} with its |state| set to |opened|.
```
mounirlamouri: Wouldn't be simpler/shorter to simply instantiate the object and say in PIPWindow spec part that when instantiated, its state is "opened"?

index.bs
```diff
@@ -178,13 +178,10 @@ the user agent MUST run the following steps:
...
     these steps.
-2. Let |pipWindow| be the current Picture-in-Picture window.
-3. Let |pipVideo| be the video contained in the |pipWindow|.
-4. Let |video| be the source video of |pipVideo| frames.
-5. Close |pipWindow|.
-6. Render |pipVideo| frames in the |video|.
-7. Unset {{pictureInPictureElement}}.
-8. <a>Queue a task</a> to <a>fire an event</a> with the name
+2. Set the <a>Picture-in-Picture window</a> associated with
+    {{pictureInPictureElement}} |state| to |closed|.
```
mounirlamouri: What about creating something called "close" and then say that when a PIPWindow is closed, its state switches to close. I think it will have the benefit of keeping "state" internal and allowing to easily expand what "close" means.

index.bs
```diff
@@ -325,8 +321,16 @@ interface PictureInPictureWindow {
 };
 </pre>
 
-The {{PictureInPictureWindow}} is a simple proxy to get information such as
-dimensions about the current Picture-in-Picture window.
+A {{PictureInPictureWindow}} instance represents a <a>Picture-in-Picture
+window</a>. It has a |state| that is either |opened| or |closed|.
```
mounirlamouri: To elaborate on what's above, you could say that "when instantiated, an instance of PIPWindow has its state set to opened"

index.bs
```diff
@@ -325,8 +321,16 @@ interface PictureInPictureWindow {
...
+
```
mounirlamouri: You should probably mention that a PIPWindow is associated to an element?

index.bs
```diff
@@ -325,8 +321,16 @@ interface PictureInPictureWindow {
...
+The {{width}} attribute is the width in <a lt=px value>CSS pixels</a> of the
+<a>Picture-in-Picture window</a> associated with {{pictureInPictureElement}}.
```
mounirlamouri: What about "The width attribute MUST return the width of the window in CSS pixels if the state is opened. Otherwise, it MUST return 0."

Note that I'm not sure how it's phrased in the CSS  spec for the "window width" so maybe copy some phrasing from there.

index.bs

index.bs

index.bs

index.bs

index.bs

beaufortfrancois:
@mounirlamouri Hopefully this is better now.

index.bs
```diff
@@ -325,8 +319,20 @@ interface PictureInPictureWindow {
 };
 </pre>
 
-The {{PictureInPictureWindow}} is a simple proxy to get information such as
-dimensions about the current Picture-in-Picture window.
+A {{PictureInPictureWindow}} instance represents a <a>Picture-in-Picture
+window</a> associated with an {{Element}}. When instantiated, an instance of
```
mounirlamouri: Could it be `{{HTMLVideoElement}}` instead of `{Element}`? I mean, the spec today only mention videos, right?

index.bs
```diff
@@ -325,8 +319,20 @@ interface PictureInPictureWindow {
...
+{{PictureInPictureWindow}} has its |state| set to |opened|.
+
+When the <dfn>close window algorithm</dfn> with an instance of
+{{PictureInPictureWindow}} is invoked, its |state| is set to |closed|.
+
+The {{width}} attribute MUST return the width in <a lt=px value>CSS pixels</a> of the
+<a>Picture-in-Picture window</a> associated with {{pictureInPictureElement}} if
+the |state| is |opened|. Otherwise, it MUST return 0.
+
+The {{height}} attribute is the height in <a lt=px value>CSS pixels</a> of the
```
mounirlamouri: The wording here is different from above. Can you update it?

beaufortfrancois: Oops. Thanks!

index.bs
beaufortfrancois: Done. Thanks!

beaufortfrancois:
@mounirlamouri I think we're good now.

PR #73, opened 2018-06-01
Do not pause playback when exiting Picture-in-Picture

beaufortfrancois:
Following Safari's implementation, video playback MUST not pause when exiting Picture-in-Picture from the API.

@mounirlamouri WDYT of this simple change?

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/73.html" title="Last updated on Jun 12, 2018, 5:19 AM GMT (fbd8519)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/73/2ed9c31...fbd8519.html" title="Last updated on Jun 12, 2018, 5:19 AM GMT (fbd8519)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -192,9 +192,9 @@ the user agent MUST run the following steps:
 
 It is NOT RECOMMENDED that video playback state changes when the <a>exit
 Picture-in-Picture algorithm</a> is invoked. The website SHOULD be in control
-of the experience if it's website's initiated. User agent MAY expose
+of the experience if it is website initiated. However, user agent MAY expose
 Picture-in-Picture window controls that change video playback state (e.g.
-pause) though.
+pause).
 
 ## Disable Picture-in-Picture ## {#disable-pip}
 
--- a/index.bs
+++ b/index.bs
@@ -190,6 +190,9 @@ the user agent MUST run the following steps:
     {{leavepictureinpicture}} at the |video| with its {{bubbles}} attribute
     initialized to true.
 
+The user agent SHOULD NOT change video playback state when video leaves
+Picture-in-Picture.
+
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
(gentle ping) @mounirlamouri

index.bs
```diff
@@ -190,6 +190,9 @@ the user agent MUST run the following steps:
     {{leavepictureinpicture}} at the |video| with its {{bubbles}} attribute
     initialized to true.
 
+The user agent SHOULD NOT change video playback state when video leaves
+Picture-in-Picture.
```
mounirlamouri: I think this should only apply when the `exitPictureInPicture()` method is called. As defined here, it would apply to *all* ways of leaving Picture-in-Picture, right?

beaufortfrancois: I can be more specific and say something like "The user agent SHOULD NOT change video playback state when the exit Picture-in-Picture algorithm is run".
This means it would apply only to when the `exitPictureInPicture()` method is called and when `disablePictureInPicture` is set on an existing PiP video. What do you think?

mounirlamouri: That SGTM. Though, I would consider using NOT RECOMMENDED just because technically, changing the playback state wouldn't follow the algorithm already. It may be interesting to give more info, explaining that it may be okay to have side effects from interacting with a UI element but if it's website initiated, the website should be in control of the experience.

beaufortfrancois: What do you think of https://github.com/WICG/picture-in-picture/pull/73/commits/f0c6e707b7d032b0248e711054cee2e9381c3942?

PR #82, opened 2018-09-12
Add "Interaction with Page Visibility" section

### LINKED ISSUE #80: Interaction with page visibility
It is not entirely clear how the picture-in-picture API interacts with the [page visibility API](https://www.w3.org/TR/page-visibility/) when the picture-in-picture video is visible but the browser itself is minimized:

Page visibility API, [visibilityState attribute](https://www.w3.org/TR/page-visibility/#visibilitystate-attribute) defines (emphasis added):
> On getting, the visibilityState attribute the user agent must run the steps to determine the visibility state:
> 1. Let doc be the Document of the top level browsing context.
> 2. If the defaultView of doc is null, return hidden.
> 3. Otherwise, return the VisibilityState value that best matches the visibility state of doc:
>    1. Return "visible" if:
>       1. The user agent is not minimized and doc is the foreground tab.
>       2. The user agent is fully obscured by an accessibility tool, like a magnifier, but a view of the doc is shown.
>    2. **Return "hidden" if:**
>       1. **The user agent is minimized.**
>       2. The user agent is not minimized, but doc is on a background tab.
>       3. The user agent is to unload doc.
>       4. The Operating System lock screen is shown.

From this section it looks like visibilityState should be "hidden" in such a situation.

However, the [enum definition](https://www.w3.org/TR/page-visibility/#dom-visibilitystate-visible) says (emphasis added):
> visible
> The Document is **at least partially visible** on **at least one screen**. This is the same condition under which the hidden attribute is set to false.

From this section it looks like visibilityState should be "visible" in such a situation.

It would probably be a good idea to specify the interaction with page visibility explicitly in the picture-in-picture API.

beaufortfrancois:
FIX #80

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/82.html" title="Last updated on Oct 24, 2018, 1:50 PM GMT (44db90c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/82/6dca358...44db90c.html" title="Last updated on Oct 24, 2018, 1:50 PM GMT (44db90c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -238,10 +238,9 @@ available controls on the Picture-in-Picture window.
 
 The [[Page-Visibility]] specification defines a {{Document/visibilityState}}
 attribute used to determine the visibility state of a top level browsing
-context. For the purpose of Picture-in-Picture, the
-{{Document/visibilityState}} attribute is always "visible" when
-{{pictureInPictureElement}} is set and the Operating System lock screen is not
-shown.
+context. For the purpose of Picture-in-Picture, the {{Document/visibilityState}}
+MUST always return "visible" when {{pictureInPictureElement}}
+is set and the Picture-in-Picture window is visible.
 
 ## One Picture-in-Picture window ## {#one-pip-window}
 
--- a/index.bs
+++ b/index.bs
@@ -240,7 +240,8 @@ The [[Page-Visibility]] specification defines a {{Document/visibilityState}}
 attribute used to determine the visibility state of a top level browsing
 context. For the purpose of Picture-in-Picture, the
 {{Document/visibilityState}} attribute is always "visible" when
-{{pictureInPictureElement}} is set.
+{{pictureInPictureElement}} is set and the Operating System lock screen is not
+shown.
 
 ## One Picture-in-Picture window ## {#one-pip-window}
 
--- a/index.bs
+++ b/index.bs
@@ -34,6 +34,8 @@ spec: Remote-Playback; urlPrefix: https://w3c.github.io/remote-playback/#dfn-
     type: dfn
         text: local playback device
         text: local playback state
+spec: Page-Visibility; urlPrefix: https://www.w3.org/TR/page-visibility/
+    type: attribute; text: visibilityState; for: Document; url: dom-visibilitystate
 </pre>
 
 <pre class="link-defaults">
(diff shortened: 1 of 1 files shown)


ewoerner:
With the PR, the following case is treated as "visible", which is probably not intended:
1. video is in PiP mode and
2. lock screen is active

beaufortfrancois:
Indeed! I've updated PR.

ewoerner:
@beaufortfrancois looks good to me now, thank you.

beaufortfrancois:
@mounirlamouri I've tried Safari and it looks like Picture-in-Picture doesn't make any difference.

@jernoble, is that intended? Shall we set `visibilityState` attribute to be "visible" when
`pictureInPictureElement` is set and the Operating System lock screen is not
shown even when document is in background.

mounirlamouri:
To summarise a discussion @beaufortfrancois and I had offline: Chrome Android auto-PIP exposes that the page is hidden but Chrome desktop PIP feature seems to make the page visible by default. This isn't intentional and I'm actually surprised it does. I do understand some of the concerns shared from the bug report but my understanding is that these concerns are mostly for backward compatibility and websites expecting PIP should be using the API discover the state. I do not have a strong opinion either way but compatibility matters here.

beaufortfrancois:
For info, you can play with PiP and Page Visibility API at https://beaufortfrancois.github.io/sandbox/media/picture-in-picture-playground

@mounirlamouri I wasn't able to see `document.visibilityState` set to `hidden` when video plays AutoPiP in Chrome for Android in Canary. Did I miss anything?

mounirlamouri:
@beaufortfrancois I may have been very wrong :)

Though, the compatibility point still applies but between Safari and Chrome I guess.

beaufortfrancois:
After talking to @jernoble at [TPAC 2018](https://www.w3.org/2018/10/TPAC/), we agreed we should go ahead with this PR.

@mounirlamouri We'd have to fix Chrome for Android implementation of AutoPiP to make sure that `document.visibilityState` is `visible` when video is in Picture-in-Picture.

mounirlamouri:
Safari would also need to update its implementation, right?

jernoble:
Yes, and we'd like to.

index.bs
```diff
@@ -232,6 +234,15 @@ Picture-in-Picture.
...
 available controls on the Picture-in-Picture window.
 
+## Interaction with Page Visibility ## {#page-visibility}
+
+The [[Page-Visibility]] specification defines a {{Document/visibilityState}}
+attribute used to determine the visibility state of a top level browsing
+context. For the purpose of Picture-in-Picture, the
+{{Document/visibilityState}} attribute is always "visible" when
+{{pictureInPictureElement}} is set and the Operating System lock screen is not
+shown.
```
mounirlamouri: Let's rephrase to:
```
The {{Document/visibilityState}} MUST always return "visible" when {{pictureInPictureElement}} is set and the Picture-in-Picture window is visible.
```

beaufortfrancois: Done

beaufortfrancois:
FYI I've filed https://bugs.chromium.org/p/chromium/issues/detail?id=898495 to make sure Chromium implementation is fixed.

PR #90, opened 2018-10-18
Add Web Platform Tests badges

beaufortfrancois:


beaufortfrancois:
Note that I'll update this to use the official wpt badge API when it lands: https://github.com/web-platform-tests/wpt.fyi/pull/674#issuecomment-431018715

PR #95, opened 2018-10-24
Add maximum picture-in-picture window size

beaufortfrancois:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/95.html" title="Last updated on Oct 24, 2018, 7:46 AM GMT (e4fc541)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/95/d7b0f01...e4fc541.html" title="Last updated on Oct 24, 2018, 7:46 AM GMT (e4fc541)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -203,6 +203,9 @@ Styles applied to |video| (such as opacity, visibility, transform, etc.) MUST
 NOT apply in the Picture-in-Picture window. Its aspect ratio is based on the
 video size.
 
+It is also RECOMMENDED that the Picture-in-Picture window has a maximum size
+set to half of the screen.
+
 ## Exit Picture-in-Picture ## {#exit-pip}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -203,6 +203,9 @@ Styles applied to |video| (such as opacity, visibility, transform, etc.) MUST
 NOT apply in the Picture-in-Picture window. Its aspect ratio is based on the
 video size.
 
+It is also RECOMMENDED that the Picture-in-Picture window has a maximum size
+set to half of the screen.
```
mounirlamouri: I think we should say maximum and minimum and give examples. Something like:
"It is also RECOMMENDED that the Picture-in-Picture window has a maximum and minimum size. For example, it could be restricted to be between a quarter and a half of one dimension of the screen."

PR #102, opened 2018-11-26
Be more explicit about PiP window visibility

beaufortfrancois:
@yoavweiss noted that spec doesn't tell much about the obvious visibility state of the Picture-in-Picture window. This PR fixes this.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/102.html" title="Last updated on Nov 26, 2018, 11:36 AM GMT (da07e5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/102/ed25bd1...da07e5e.html" title="Last updated on Nov 26, 2018, 11:36 AM GMT (da07e5e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -54,7 +54,8 @@ spec:html; type:attribute; for:HTMLMediaElement; text:readyState
 Many users want to continue consuming media while they interact with other
 content, sites, or applications on their device. A common UI affordance for
 this type of activity is Picture-in-Picture (PiP), where the video is contained
-in a separate miniature window that is always on top of other windows.
+in a separate miniature window that is always on top of other windows. This
+window stays visible even when user agent is not visible.
 Picture-in-Picture is a common platform-level feature among desktop and mobile
 OSs.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -267,11 +268,14 @@ available controls on the Picture-in-Picture window.
...
+
 The [[Page-Visibility]] specification defines a {{Document/visibilityState}}
-attribute used to determine the visibility state of a top level browsing
-context. For the purpose of Picture-in-Picture, the {{Document/visibilityState}}
-MUST always return "visible" when {{pictureInPictureElement}}
-is set and the Picture-in-Picture window is visible.
+attribute used to determine the visibility state of a <a>top-level browsing
+context</a>. For the purpose of Picture-in-Picture, the
+{{Document/visibilityState}} MUST always return "visible" when
+{{pictureInPictureElement}} is set.
```
yoavweiss: Looks good from the visibility state perspective. It's not clear in the spec which mechanisms users will typically have to eventually close such web sites, and how those mechanisms interact with the "exit pip" algorithm. Would be good to add some wording on that front (maybe a note would be enough).

yoavweiss: Yeah, just to say that browsers should make sure there is a way for users to eventually close that window (and it's not a way for annoying sites/ads to remain in the user's face forever)

PR #105, opened 2018-11-26
Add Secure Context

beaufortfrancois:
@yoavweiss suggested adding one section for Secure Context that explains why API is not restricted to Secure Context.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/105.html" title="Last updated on Nov 26, 2018, 12:15 PM GMT (2456bcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/105/ed25bd1...2456bcd.html" title="Last updated on Nov 26, 2018, 12:15 PM GMT (2456bcd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -45,6 +45,7 @@ spec:dom; type:attribute; text:bubbles
 spec:dom; type:dfn; for:NamedNodeMap; text:element
 spec:dom; type:interface; text:Document
 spec:html; type:attribute; for:HTMLMediaElement; text:readyState
+spec:html; type:dfn; for:/; text:browsing context
 </pre>
 
 # Introduction # {#intro}
@@ -428,6 +429,12 @@ The <a>task source</a> for all the tasks queued in this specification is the
 The API applies only to {{HTMLVideoElement}} in order to start on a minimal
 viable product that has limited security issues. Later versions of this
 specification may allow PIP-ing arbitrary HTML content.
+
+## Secure Context ## {#secure-context}
+
+The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
+to web applications that user agents usually offer natively on all media
+regardless of the <a>browsing context</a>.
 
 ## Feature Policy ## {#feature-policy}
 


index.bs
```diff
@@ -429,6 +430,12 @@ The API applies only to {{HTMLVideoElement}} in order to start on a minimal
 viable product that has limited security issues. Later versions of this
 specification may allow PIP-ing arbitrary HTML content.
 
+## Secure Context ## {#secure-context}
+
+The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
```
yoavweiss: Maybe "The API is not limited to SECURE-CONTEXTs because it is not likely to suffer from compatibility issues with existing MITM proxies, does expose the user to increased risks in case of content injection by an MITM proxy and does not expose any information about the user in case of passive monitoring of the user's network."?

mounirlamouri: Isn't that a bit too extensive? I mean, I think the current description may benefit from more details but that sounds a bit too much. We could just mention that the feature, in addition of being built on top of features that are not limited to Secure Contexts isn't "adding any capabilities that can be a privacy or security risks".

beaufortfrancois:
@mounirlamouri What's your call?

index.bs
marcoscaceres: Why does this need to be justified at all? I'd remove this justification - it's not needed.

mounirlamouri:
This LGTM. I realise that @marcoscaceres suggested to drop the section but I think some folks feel very strongly about everything being limited to secure contexts so I think it makes sense to explain why we did what we did.

marcoscaceres:
Oh whoopsy, I totally misread it! I thought it was secure context! 

Ok, yeah, please just make it SecureContext... if we implement this in Gecko, we will undoubtedly make it SecureContext. 

Maybe we can get a second opinion from someone at Apple?

beaufortfrancois:
@jernoble Do you have an opinion on whether we should restrict Picture-in-Picture Web API to secure contexts only?

mounirlamouri:
> @jernoble Do you have an opinion on whether we should restrict Picture-in-Picture Web API to secure contexts only?

I'm not @jernoble but if Gecko wants to make this Secure Context, it's fine. We could add a note that some UA can do that but the arguments we have in the spec IMO still apply. Last I checked, Safari didn't restrict AirPlay to Secure Context.

marcoscaceres:
Gecko will go with whatever has consensus amongst implementers, but ultimately we'd like to put new APIs in `SecureContexts`. If it doesn't hurt site devs, and there is not real strong case to *not* add it, then I'd really encourage you to add it sooner rather than later 🙏 ... otherwise, it makes a mess later (see Geolocation and Gamepad API).

mounirlamouri:
Do we think Fullscreen should be limited to secure contexts?

marcoscaceres:
On the Mozilla side, we'd like to eventually have everything on the platform run in secure contexts. However, how we do that with older APIs, like fullscreen, may require a different approach than simply adding some IDL.

jernoble:
Yeah, for us, that horse has already left the barn. Picture-in-Picture API works in non-secure contexts, and it's part of the default UA controls for `<video>` elements, so the feature itself even works for local files, etc.  In addition, it's not an especially "powerful feature" that would have additional reasons to gate behind a secure context check.

All that said, an arbitrary element, fully interactive picture-in-picture API would definitely count as a "powerful feature", and for anti-phishing purposes alone should really be gated behind a secure context check. Fullscreen API, too, though that falls into the same caveat of "horse has left the barn".

marcoscaceres:
>  and it's part of the default UA controls for `<video>` elements

I agree that users should be able to access the UA controls irrespective of it being a secure context.    

> In addition, it's not an especially "powerful feature" that would have additional reasons to gate behind a secure context check.

Agree.

PR #106, opened 2018-11-26
Exit pip on document unload

beaufortfrancois:
It wasn't clear to @yoavweiss that the exit pip algorithm was run on document unload. This PR addresses this.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/106.html" title="Last updated on Nov 26, 2018, 12:30 PM GMT (08960b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/106/ed25bd1...08960b9.html" title="Last updated on Nov 26, 2018, 12:30 PM GMT (08960b9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -230,6 +230,9 @@ of the experience if it is website initiated. However, user agent MAY expose
 Picture-in-Picture window controls that change video playback state (e.g.
 pause).
 
+As one of the <a>unloading document cleanup steps</a>, run the <a>exit
+Picture-in-Picture algorithm</a>.
+
 ## Disable Picture-in-Picture ## {#disable-pip}
 
 Some pages may want to disable Picture-in-Picture for a video element. To


index.bs
```diff
@@ -230,6 +230,9 @@ of the experience if it is website initiated. However, user agent MAY expose
 Picture-in-Picture window controls that change video playback state (e.g.
 pause).
 
+As one of the <a>unloading document cleanup steps</a>, run the <a>exit
+Picture-in-Picture algorithm</a>.
```
yoavweiss: It would be significantly better to actually add that step to the algorithm in HTML, and only include a note here saying that this is called from that point

mounirlamouri: It sounds that these steps were designed to have API hook into them, right? Is it really worth having part of the Picture-in-Picture API behoviour defined in HTML?

beaufortfrancois: For info, the fullscreen API hooks the same way into it: https://fullscreen.spec.whatwg.org/#model (look for `unloading document cleanup steps`). @yoavweiss WDYT?

yoavweiss: OK, fair enough

PR #112, opened 2018-12-20
Support "Auto Picture-in-Picture"

### LINKED ISSUE #111: Support "Auto Picture-in-Picture"
Video meetings web apps would benefit from automatic Picture-in-Picture behavior when user switches back and forth between web app and other applications/tabs. This is currently not possible with the user gesture requirement in https://wicg.github.io/picture-in-picture/#request-pip (step 6).

After some [discussion](https://github.com/WICG/picture-in-picture/pull/109), we think a declarative behaviour may be the best we can can propose as it allows browsers to correctly implement an “auto-pip on visibility change” feature without opening up a security or annoyance can of worms. Here's what I'm envisioning.

_Note: I’m using the `autopictureinpicture` keyword name. We can use comments to be pick a better name later if needed._

### IDL
```js
partial interface HTMLVideoElement {
  attribute boolean autoPictureInPicture;
};
```

### Javascript
```js
video.srcObject = await navigator.mediaDevices.getUserMedia({ video: true });
video.autoPictureInPicture = true; // <- This is where the magic happens.

video.addEventListener('enterpictureinpicture', function(event) {
  // Video entered Picture-in-Picture (potentially automatically)
});
video.addEventListener('leavepictureinpicture', function(event) {
  // Video left Picture-in-Picture (potentially automatically)
});
```

When `autoPictureInPicture` is set to `true`,  video will attempt to enter Picture-in-Picture automatically (without user gesture) when page becomes hidden. And existing Picture-in-Picture window will be closed automatically when page becomes visible. We'll use [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) events for this.

I believe this proposal is simple and concise but there is something to consider.
What if we have two video elements with `autoPictureInPicture` (or more). Spec could say that the first video element matching `document.querySelector('video[autopictureinpicture]')` will always get the “auto-pip on visibility change” behaviour.
But what if the video element is not attached to the DOM? In that case, we would have to monitor all `autoPictureInPicture` values set, potentially managing a stack of video elements allowed to use this feature. Eventually, this would have to be defined in the spec. In the end, [there can be only one](https://www.youtube.com/watch?v=sqcLjcSloXs).

Note that I’ve considered alternatives such as a `document.autoPictureInPictureElement` attribute that web developers could set. However I’d prefer using a [HTML content attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Content_versus_IDL_attributes) as this doesn't require web developers to use Javascript for such a simple opt-in thing. And it also pollutes the `document` element.

What do you think?

beaufortfrancois:
FIX #111


jernoble:
> "The autoPictureInPictureElement is the latest video element with the autoPictureInPicture attribute set."

I'm entirely sure what "the latest video element" here means.  Is it the most recently created or parsed video element? Is it the video element whose autoPictureInPicture attribute was set most recently?  Is it the one who most recently started playing?

beaufortfrancois:
> > "The autoPictureInPictureElement is the latest video element with the autoPictureInPicture attribute set."
> 
> I'm [edit: not] entirely sure what "the latest video element" here means. Is it the most recently created or parsed video element? Is it the video element whose autoPictureInPicture attribute was set most recently? Is it the one who most recently started playing?

You're right. It wasn't clear enough. I've updated it. 
It is the video element whose `autoPictureInPicture` attribute was set most recently.

juberti:
This all looks sensible to me. Thanks!

index.bs
```diff
@@ -259,6 +261,38 @@ the user agent SHOULD run these steps:
 2. If |video| is {{pictureInPictureElement}}, run the <a>exit
     Picture-in-Picture algorithm</a>.
 
+## Auto Picture-in-Picture ## {#auto-pip}
+
+Some pages may want to automatically enter and leave Picture-in-Picture for
+a video element; for example, video meetings web apps would benefit from
+some automatic Picture-in-Picture behavior when user switches back and forth
+between web app and other applications/tabs.
```
mounirlamouri: "between _the_ web app"?

index.bs
```diff
@@ -259,6 +261,38 @@ the user agent SHOULD run these steps:
...
+To support these use cases, a new {{autoPictureInPicture}} attribute is
+added to the list of content attributes for video elements.
```
mounirlamouri: You could mark this paragraph as non-normative.

index.bs
```diff
@@ -259,6 +261,38 @@ the user agent SHOULD run these steps:
...
+
+A corresponding {{autoPictureInPicture}} IDL attribute which reflects the
+value of element’s {{autoPictureInPicture}} content attribute is added to
+the {{HTMLVideoElement}} interface. The {{autoPictureInPicture}} IDL attribute
+MUST <a>reflect</a> the content attribute of the same name.
```
mounirlamouri: Sounds like the two sentences say the same thing, don't they?

index.bs
```diff
@@ -259,6 +261,38 @@ the user agent SHOULD run these steps:
...
+
+The <dfn>autoPictureInPictureElement</dfn> is the video element whose
+{{autoPictureInPicture}} attribute was set most recently.
```
mounirlamouri: What happens if this element loses the attribute?

index.bs
```diff
@@ -259,6 +261,38 @@ the user agent SHOULD run these steps:
...
+2. Let |video| be <a>autoPictureInPictureElement</a>.
+3. Let |userActivationRequired| be `false`.
+4. Run the <a>request Picture-in-Picture algorithm</a> with |video| and
+    |userActivationRequired|.
+
+When the visibility state of a <a>top-level browsing context</a> value
+switches from "visible" to "hidden", the user agent MAY run these steps:
+
+1. If <a>autoPictureInPictureElement</a> is null, abort these steps.
+2. Run the <a>exit Picture-in-Picture algorithm</a>.
```
mounirlamouri: We should also check that the `pictureInPictureElement` is `autoPictureInPictureElement`

index.bs
beaufortfrancois: Nothing.
Do you think something should happen?

index.bs
beaufortfrancois: You're right. Thanks!

index.bs
beaufortfrancois: I'm not sure to be honest. It is the same as https://w3c.github.io/remote-playback/#the-disableremoteplayback-attribute.
Which one sentence should I pick then?

index.bs
mounirlamouri: I guess "was set most recently" may be too vague. If I do: `video.setAttribute('autopictureinpicture', ''); videos.setAttribute('autopictureinpicture', undefined);`, `video` had the attribute set most recently but the page clearly said that it did not want to have it go in auto-pip anymore.

index.bs
```diff
@@ -259,6 +261,42 @@ the user agent SHOULD run these steps:
...
+the {{HTMLVideoElement}} interface. The {{autoPictureInPicture}} IDL attribute
+MUST <a>reflect</a> the content attribute of the same name.
+
+The <dfn>autoPictureInPictureElement</dfn> is the video element whose
+{{autoPictureInPicture}} attribute was set most recently.
+
+When the visibility state of a <a>top-level browsing context</a> value
+switches from "hidden" to "visible", the user agent MAY run these steps:
+
+1. If <a>autoPictureInPictureElement</a> is null, abort these steps.
```
mounirlamouri: Should we abort these steps if there is already a video in Picture-in-Picture?

index.bs
mounirlamouri: I think the second sentence is enough.

index.bs
beaufortfrancois: Would it be better?

`the video element whose {{autoPictureInPicture}} content attribute was set most recently.`.

See https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect

index.bs
beaufortfrancois: It makes sense actually. Thank you!

index.bs
beaufortfrancois: How about https://github.com/WICG/picture-in-picture/pull/112/commits/b12c3172213e67b7dab8e7b88b09974671db2b42

beaufortfrancois:
@mounirlamouri I've added the "potentially playing" restriction we've discussed about offline.

index.bs
```diff
@@ -259,6 +262,46 @@ the user agent SHOULD run these steps:
...
+between the web app and other applications/tabs.
+To support these use cases, a new {{autoPictureInPicture}} attribute is
+added to the list of content attributes for video elements.
+
+The {{autoPictureInPicture}} IDL attribute MUST <a>reflect</a> the content
+attribute of the same name.
+
+The <dfn>autoPictureInPictureElement</dfn> is the video element, among all
+video elements with the {{autoPictureInPicture}} attribute currently set,
+whose {{autoPictureInPicture}} attribute was set most recently. Note that the
```
mounirlamouri: I think _was set most recently_ would benefit from a clear definition. Happy to have this dealt with in a follow-up though.

scottlow:
@beaufortfrancois I'm late to the party, but this looks good to me. Thanks!

beaufortfrancois:
Good to hear @scottlow!

index.bs
beaufortfrancois: It is tracked at https://github.com/WICG/picture-in-picture/issues/115

beaufortfrancois:
@mounirlamouri I've filed https://github.com/WICG/picture-in-picture/issues/115 and replaced "potentially playing" with "paused" at https://github.com/WICG/picture-in-picture/pull/112/commits/3b17c21db5110d8127dd37511e005589fae46afa.
If that still looks good to you, feel free to merge.

PR #114, opened 2019-01-21
Treat DocumentOrShadowRoot as an interface mixin

foolip:
Matching https://dom.spec.whatwg.org/#mixin-documentorshadowroot.

***
<a href="https://pr-preview.s3.amazonaws.com/foolip/picture-in-picture/pull/114.html" title="Last updated on Jan 21, 2019, 2:07 PM UTC (03da806)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/114/658c859...foolip:03da806.html" title="Last updated on Jan 21, 2019, 2:07 PM UTC (03da806)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -355,7 +355,7 @@ parallel</a>:
 ## Extension to <code>DocumentOrShadowRoot</code> ## {#documentorshadowroot-extension}
 
 <xmp class="idl">
-partial interface DocumentOrShadowRoot {
+partial interface mixin DocumentOrShadowRoot {
   readonly attribute Element? pictureInPictureElement;
 };
 </xmp>


beaufortfrancois:
Thanks @foolip!

Even though it is not supported in Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=781257), do you suggest I add it to the spec?
How about Web Platform Tests? Are they going to complain?

foolip:
Partial interfaces, the old [NoInterfaceObject] + `A implements B` pattern and the new interface mixin + `A includes B` pattern and are just a spec writing convenience, doing exactly the same thing as if everything was defined in a single interface definition.

So no tests of behavior would be affected by this. idlharness.js tests already handle mixins. After the spec is updated, a PR against wpt will be automatically be created, I'll try to remember to mention you there so you can see how it works.

beaufortfrancois:
I'll be monitoring these changes. Thanks!

FYI @mounirlamouri

foolip:
The change reached wpt in https://github.com/web-platform-tests/wpt/pull/14990.

PR #118, opened 2019-02-08
Some editorial changes

chrisn:
I've just had a quick read through the PiP spec, looks great so far! I made a few editorial changes as I was reading. The main one is that some text and the examples used the term "Picture-in-Picture mode", but elsewhere this is simply "Picture-in-Picture", so I edited to use "Picture-in-Picture mode" throughout. I hope these changes are OK with you.

***
<a href="https://pr-preview.s3.amazonaws.com/chrisn/picture-in-picture/pull/118.html" title="Last updated on Feb 8, 2019, 6:08 PM UTC (81a101c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/118/b926a27...chrisn:81a101c.html" title="Last updated on Feb 8, 2019, 6:08 PM UTC (81a101c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -62,16 +62,16 @@ OSs.
 This specification aims to allow websites to initiate and control this behavior
 by exposing the following sets of properties to the API:
 
-* Notify the website when it enters and leave Picture-in-Picture mode.
+* Notify the website when it enters and leaves Picture-in-Picture mode.
 * Allow the website to trigger Picture-in-Picture via a user gesture on a video
     element.
-* Allow the website to know the size of Picture-in-Picture window and notify
+* Allow the website to know the size of the Picture-in-Picture window and notify
     the website when it changes.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -134,6 +140,7 @@ extensible.
 
 <script>
   let pipWindow;
+  const pipButton = document.getElementById('pipButton');
```
beaufortfrancois: Shall we add `const video = document.getElementById('video');`as well then?

chrisn:
Many thanks, @beaufortfrancois. You're right, will do.

index.bs
chrisn: Sorry, I missed that... I just pushed a fix, please take a look!

beaufortfrancois:
Thank you again @chrisn!

PR #123, opened 2019-04-16
Add explainer for v2 (arbitrary content)

beccahughes:
I wrote an explainer for the arbitrary content based on @mounirlamouri's proposal document. I was thinking we should keep all the arbitrary content bits on a separate `v2` branch so we can iterate on it easily and fix issues and then when are ready we can merge it into master.

@beaufortfrancois @mounirlamouri WDYT?

### DIFF
--- a/v2_explainer.md
+++ b/v2_explainer.md
@@ -1,0 +1,85 @@
+# Picture-in-Picture V2 Explained
+
+## Background: Picture-in-Picture V1
+
+Picture-in-Picture launched for HTMLVideoElement early in 2019. The API was
+designed to be extensible, like the Fullscreen API but at the moment is only
+defined for video elements. The implementation takes the video layer and moves
+it into a window that follows display rules specific to Picture-in-Picture
+(always on top, fixed aspect ratio, etc.).
+
+## What's Missing from V1?
+
+Various websites expressed frustration around the limitations of the API. The
+feature requests fell into two categories: customising the look & feel and
+customising the user experience. The former can technically be done using canvas
+elements but the latter requires a different model that would require user
+interaction and therefore no longer be compatible with Android, iOS and macOS
+APIs.
+
+## Goals
+
+  1. **Allow sites to Picture-in-Picture arbitary content**
+  2. **A site should be able to choose whether that content is interactive.** If
+the platform (e.g. Android) does not support interactive content then the site
+should be able to easily fallback to non-interactive content
+
+## Non-Goals
+
+The Picture-in-Picture window is not a popup. Therefore, the content should not
+live beyond the lifetime of the page.
+
+## Sample Usage
+
+Here's how a page might request Picture-in-Picture for it's media controls and
+then move the video element to the new window. When the window is closed it will
+then move the video element back to the page.
+
+```html
+<video id=video>
+<div id=controls>
+```
+
+```js
+const controls = document.getElementById('controls');
+const video = document.getElementById('video');
+
+const pipWindow = await controls.requestPictureInPicture({
+  aspectRatio: 1,
+  interactive: true
+});
+
+pipWindow.document.body.appendChild(video);
+
+pipWindow.addEventListener('leavepictureinpicture', () => {
+  document.body.appendChild(video);
+});
+```
+
+The `aspectRatio` will be used to determine the size of the Picture-in-Picture
+window.
+
+## Alternatives Considered
+
+We considered using window.open with additional `picture-in-picture` and
+`interactive` options. However to expose that `interactive` or
+`picture-in-picture` is not supported in `window.open()` would require us
+to change this API to throw an additional exception which could cause
+compatability issues. The new Picture-in-Picture window also does not have
+a URL whereas `window.open()` has an argument for a URL so this may cause
+confusion. Finally, `window.open`'s window also has a different lifetime
+than the Picture-in-Picture window.
+
+We also considered providing a URL to a document level API. However, this
+causes security concerns as well as having a poor user experience due to having
+to load the URL.
+
+## Open Questions
+
+The new `requestPictureInPicture` API has different parameters and a different algorithm
+so we should consider renaming it
+([#129] (https://github.com/WICG/picture-in-picture/issues/129)).
+
+# Acknowledgements
+
+mounirlamouri@ wrote the original version of this proposal.


mounirlamouri:
It might be good to more clearly state goals and non-goals as explained in https://github.com/w3ctag/w3ctag.github.io/blob/master/explainers.md

Some recommend to drop IDL in explainers. I do not have a strong opinion but I do not think we should go with the Document *and* Element level API. Dropping the IDL from the explainer will punt that discussion ;)

beccahughes:
Okay done

v2_explainer.md
```diff
@@ -0,0 +1,76 @@
...
+live beyond the lifetime of the page.
+
+## Sample Usage
+
+Here's how a page might request Picture-in-Picture for it's media controls and
+then move the video element to the new window. When the window is closed it will
+then move the video element back to the page.
+
+```html
+<video id=video>
```
beaufortfrancois: I naively thoughts most people would simply request PiP for the entire video player element and have CSS media queries to customize it.

```html
<div id="video-player">
  <div id="video-controls">...</div>
  <video id="video">...</video>
</div>
```
```js
const player = document.querySelector('#video-player');
const pipWindow = await player.requestPictureInPicture({
  aspectRatio: 1,
  width: 200,
  height: 200,
  interactive: true
});
```

v2_explainer.md
```diff
@@ -0,0 +1,76 @@
...
+});
+```
+
+Both `width` and `height` will only be used as hints and could be ignored by the
+system but `aspectRatio` will be kept the same.
+
+## Alternatives Considered
+
+We considered using window.open with an additional `picture-in-picture` option.
+However, there are limitations in `window.open()` and extending it will probably
```
beaufortfrancois: Do you mind explain why?

beccahughes: I have added it.

v2_explainer.md
beccahughes: The `requestPictureInPicture` copies the DOM element to the new page so it would break video playback. Having an exception to move elements in a copy algorithm might be a bit confusing. We could move the DOM element however we would then not have an element to set for `pictureInPictureElement`.

Since the new window is also a new document it would have its own set of stylesheets and therefore any styles would also need to be copied to the PiP window. We cannot use CSS media queries here.

beaufortfrancois: Thinking out loud, shall we rename this method to `copyToPictureInPictureWindow` then so that it's clear what it's doing?

beccahughes: I think we should rename this because `requestPictureInPicture` on the video element has different parameters and a different underlying algorithm. It would make it clear to developers exactly what this is doing.

v2_explainer.md
```diff
@@ -0,0 +1,76 @@
...
+
+The `aspectRatio` will be used to determine the size of the Picture-in-Picture
+window.
+
+## Alternatives Considered
+
+We considered using window.open with a additional `picture-in-picture` and
+`interactive` options. However, exposing that `interactive` or
+`picture-in-picture` is not supported in `window.open()` could only return
+`null` which is not very descriptive. We could change this behaviour but it
```
beaufortfrancois: `window.open` already throws errors that could be used to detect whether or not Picture-in-Picture is not possible. See https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#document-open-steps.

beccahughes: These steps are for `document.open` whereas `window.open` can only return the window or null: https://html.spec.whatwg.org/multipage/window-object.html#window-open-steps. We would have to add support for throwing or returning an exception which sites may not expect and may cause breakage with normal `window.open` calls.

v2_explainer.md
beccahughes: I think we should resolve the naming issue outside of this PR, so I have clarified in the explainer that the name will likely change and added a tracking issue here: https://github.com/WICG/picture-in-picture/issues/129

beccahughes:
@beaufortfrancois @mounirlamouri do you have anything else you would like to add or can I merge this?

v2_explainer.md
beaufortfrancois: You're right. I was looking at the wrong algorithm.
The correct one you pointed out (https://html.spec.whatwg.org/multipage/window-object.html#window-open-steps) can also raise errors. See impl in Chrome: https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/local_dom_window.cc?type=cs&sq=package:chromium&g=0&l=1468

beaufortfrancois:
Except from the nit about `window.open`, I'm good.

v2_explainer.md
beccahughes: Ah yes I see. I have added clarification.

v2_explainer.md
```diff
@@ -0,0 +1,83 @@
...
+
+Picture-in-Picture launched for HTMLVideoElement early in 2019. The API was
+designed to be extensible, like the Fullscreen API but at the moment is only
+defined for video elements. The implementation takes the video layer and moves
+it into a window that follows display rules specific to Picture-in-Picture
+(always on top, fixed aspect ratio, etc.).
+
+## What's Missing from V1?
+
+Various partners expressed frustration around the limitations of the API. The
```
mounirlamouri: s/partners/websites/

Let's also not use "frustrations" as it's fairly negative.

v2_explainer.md
```diff
@@ -0,0 +1,83 @@
...
+<div id=controls>
+```
+
+```js
+const controls = document.getElementById('controls');
+const video = document.getElementById('video');
+
+const pipWindow = await controls.requestPictureInPicture({
+  aspectRatio: 1,
+  interactable: true
```
mounirlamouri: Why not "interactive"?

v2_explainer.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Alternatives Considered
+
+We considered using window.open with additional `picture-in-picture` and
+`interactive` options. However to expose that `interactive` or
+`picture-in-picture` is not supported in `window.open()` would require us
+to change this API to throw an additional exception which could cause
+compatability issues. The new Picture-in-Picture window also does not have
+a URL whereas `window.open()` has an argument for a URL so this may cause
+confusion.
```
mounirlamouri: `window.open()`'s window also has a different lifetime than the opener.

v2_explainer.md
beccahughes: Fixed.

v2_explainer.md
beccahughes: Fixed.

v2_explainer.md
beccahughes: Fixed.

beccahughes:
Thanks!

PR #125, opened 2019-04-17
Edit deploy script for v2 branch

beccahughes:
Updates the deploy.sh script on the v2 branch to build the HTML into the out/v2 directory

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e # Exit with nonzero exit code if anything fails
 
-SOURCE_BRANCH="master"
+SOURCE_BRANCH="v2"
 TARGET_BRANCH="gh-pages"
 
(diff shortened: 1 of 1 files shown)


deploy.sh
```diff
@@ -23,11 +23,15 @@ cd out
 git checkout $TARGET_BRANCH || git checkout --orphan $TARGET_BRANCH
 cd ..
 
+# Output should live in a directory with the branch name
+OUT_PATH="out/$SOURCE_BRANCH"
+
 # Clean out existing contents
-rm -rf out/* || exit 0
+rm -rf $OUT_PATH/*.html || exit 0
+mkdir $OUT_PATH || exit 0
```
mounirlamouri: I would consider calling `mkdir` first. Sounds that `rm` doesn't fail if the path doesn't exist so it's really a preference.

beccahughes: Done

PR #140, opened 2019-05-08
Update V2 explainer

beccahughes:
There was feedback from the TAG that it would be useful if the explainer started from the user benefit. The original explainer did a good job of this, so this copies the intro as background to the V2 explainer.

### DIFF
--- a/v2_explainer.md
+++ b/v2_explainer.md
@@ -1,5 +1,13 @@
 # Picture-in-Picture V2 Explained
 
+Many users want to continue consuming media while they interact with other
+content, sites, or applications on their device. A common UI affordance for
+this type of activity is Picture-in-Picture (PIP), where the video is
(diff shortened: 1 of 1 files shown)


chrisn:
One use case we have for the V2 functionality, that is perhaps worth mentioning, is to be able to customise the controls of our [media player](https://www.bbc.co.uk/iplayer) when in PiP mode.

beaufortfrancois:
@chrisn FYI current version of Picture-in-Picture allows you to get Next and Previous buttons in the Picture-in-Picture window thanks to the Media Session API. See https://googlechrome.github.io/samples/picture-in-picture/audio-playlist for instance.

media-playback-controls

chrisn:
Thanks, @beaufortfrancois, this looks great. To clarify, our use case is using our own visual design for such controls.

PR #149, opened 2019-06-03
Editorial: small suggestions for examples

marcoscaceres:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/pull/149.html" title="Last updated on Jun 3, 2019, 8:32 AM UTC (930ff12)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/picture-in-picture/149/f8dae9f...930ff12.html" title="Last updated on Jun 3, 2019, 8:32 AM UTC (930ff12)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -55,7 +55,7 @@ Many users want to continue consuming media while they interact with other
 content, sites, or applications on their device. A common UI affordance for
 this type of activity is Picture-in-Picture (PiP), where the video is contained
 in a separate miniature window that is always on top of other windows. This
-window stays visible even when a user agent is not visible.
+window stays visible even when the user agent is not visible.
 Picture-in-Picture is a common platform-level feature among desktop and mobile
 OSs.
 
--- a/index.bs
+++ b/index.bs
@@ -96,9 +96,11 @@ extensible.
     // If there is no element in Picture-in-Picture yet, let's request
     // Picture-in-Picture for the video, otherwise leave it.
     try {
-      !document.pictureInPictureElement
-        ? await video.requestPictureInPicture()
-        : await document.exitPictureInPicture();
+      if (document.pictureInPictureElement) {
+        await video.requestPictureInPicture()
+      } else {
+        await document.exitPictureInPicture()
+      }
     } catch (err) {
       // Video failed to enter/leave Picture-in-Picture mode.
     }
--- a/index.bs
+++ b/index.bs
@@ -55,7 +55,7 @@ Many users want to continue consuming media while they interact with other
 content, sites, or applications on their device. A common UI affordance for
 this type of activity is Picture-in-Picture (PiP), where the video is contained
 in a separate miniature window that is always on top of other windows. This
-window stays visible even when user agent is not visible.
+window stays visible even when a user agent is not visible.
 Picture-in-Picture is a common platform-level feature among desktop and mobile
 OSs.
 
@@ -92,19 +92,15 @@ extensible.
   togglePipButton.hidden = !document.pictureInPictureEnabled ||
     video.disablePictureInPicture;
 
-  togglePipButton.addEventListener('click', function() {
+  togglePipButton.addEventListener("click", async () => {
     // If there is no element in Picture-in-Picture yet, let's request
     // Picture-in-Picture for the video, otherwise leave it.
-    if (!document.pictureInPictureElement) {
-      video.requestPictureInPicture()
-      .catch(error => {
-        // Video failed to enter Picture-in-Picture mode.
-      });
-    } else {
-      document.exitPictureInPicture()
-      .catch(error => {
-        // Video failed to leave Picture-in-Picture mode.
-      });
+    try {
+      !document.pictureInPictureElement
+        ? await video.requestPictureInPicture()
+        : await document.exitPictureInPicture();
+    } catch (err) {
+      // Video failed to enter/leave Picture-in-Picture mode.
     }
   });
 </script>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -139,37 +136,40 @@ extensible.
...
     // Video entered Picture-in-Picture mode.
-    pipWindow = event.pictureInPictureWindow;
+    const { pictureInPictureWindow: pipWindow } = event;
     updateVideoSize(pipWindow.width, pipWindow.height);
-    pipWindow.addEventListener('resize', onPipWindowResize);
+    pipWindow.addEventListener("resize", onPipWindowResize);
   });
 
-  video.addEventListener('leavepictureinpicture', function() {
+  video.addEventListener("leavepictureinpicture", event => {
```
beaufortfrancois, line 157: `event` does not contain a `pictureInPictureWindow ` object. This is why we have a `pipWindow` object declared above.

index.bs
```diff
@@ -139,37 +136,40 @@ extensible.
...
-  let pipWindow;
-  const video = document.getElementById('video');
-  const pipButton = document.getElementById('pipButton');
+  const video = document.getElementById("video");
+  const pipButton = document.getElementById("pipButton");
 
-  pipButton.addEventListener('click', function() {
-    video.requestPictureInPicture()
-    .catch(error => {
+  pipButton.addEventListener("click", async () => {
```
beaufortfrancois, line 142: I'm not sure it is more readable this way.

index.bs
```diff
@@ -118,14 +114,15 @@ extensible.
...
   const video = document.getElementById('video');
 
-  video.addEventListener('enterpictureinpicture', function(event) {
+  video.addEventListener('enterpictureinpicture', event => {
     // Video entered Picture-in-Picture mode.
     const pipWindow = event.pictureInPictureWindow;
-    console.log('Picture-in-Picture window width: ' + pipWindow.width);
-    console.log('Picture-in-Picture window height: ' + pipWindow.height);
+    const { width, height } = pipWindow;
+    console.log(`Picture-in-Picture window width: ${width}`);
```
beaufortfrancois, line 121: My personal preference goes with `pipWindow.width` as it is easier to read at first glance.

index.bs
```diff
@@ -55,7 +55,7 @@ Many users want to continue consuming media while they interact with other
 content, sites, or applications on their device. A common UI affordance for
 this type of activity is Picture-in-Picture (PiP), where the video is contained
 in a separate miniature window that is always on top of other windows. This
-window stays visible even when user agent is not visible.
+window stays visible even when a user agent is not visible.
```
chrisn, line 58: should this be 'the' rather than 'a'?

index.bs
```diff
@@ -92,19 +92,15 @@ extensible.
...
-      .catch(error => {
-        // Video failed to enter Picture-in-Picture mode.
-      });
-    } else {
-      document.exitPictureInPicture()
-      .catch(error => {
-        // Video failed to leave Picture-in-Picture mode.
-      });
+    try {
+      !document.pictureInPictureElement
```
chrisn, line 99: I would recommend only using ternary expressions where the result is assigned to a variable, not purely for control flow

index.bs
marcoscaceres, line 58:
```suggestion
window stays visible even when the user agent is not visible.
```

index.bs
```diff
@@ -92,19 +92,15 @@ extensible.
...
+        ? await video.requestPictureInPicture()
+        : await document.exitPictureInPicture();
+    } catch (err) {
```
marcoscaceres, line 102:
```suggestion
    try {
      if (document.pictureInPictureElement) {
        await video.requestPictureInPicture()
      } else {
        await document.exitPictureInPicture()
      }
    } catch (err) {
```

marcoscaceres:
Unfortunately, since we moved this, I no longer have direct the ability to edit the branch... sending changes as suggestions.

index.bs
```diff
@@ -118,14 +114,15 @@ extensible.
...
 
-  video.addEventListener('enterpictureinpicture', function(event) {
+  video.addEventListener('enterpictureinpicture', event => {
     // Video entered Picture-in-Picture mode.
     const pipWindow = event.pictureInPictureWindow;
-    console.log('Picture-in-Picture window width: ' + pipWindow.width);
-    console.log('Picture-in-Picture window height: ' + pipWindow.height);
+    const { width, height } = pipWindow;
+    console.log(`Picture-in-Picture window width: ${width}`);
+    console.log(`Picture-in-Picture window height: ${height}`);
```
marcoscaceres, line 122:
```suggestion
    console.log(`Picture-in-Picture window width: ${pipWindow.width}`);
    console.log(`Picture-in-Picture window height: ${pipWindow.height}`);
```

index.bs
marcoscaceres: Doing the `.catch()` (or `.then()`) chaining is not very idiomatic given we have async/await. I'm imagining people copy/pasting this code then needing to adapt it.

index.bs
marcoscaceres, line 122:
```suggestion
    console.log(`Picture-in-Picture window width: ${pipWindow.width}`);
    console.log(`Picture-in-Picture window height: ${pipWindow.height}`);
```

index.bs
marcoscaceres: Ohh... um... ok... this seems like an odd API design choice here... why not fire an `EnterPictureInPictureEvent` (yes, it has "Enter" on the front, boo!) if the developer still needs to access `pictureInPictureWindow`? Otherwise, the API forces the developer to keep an unnecessary reference to a variable in the global or some outer scope?

index.bs
beaufortfrancois: Noted. Let's keep it this way.

index.bs
beaufortfrancois: I've filed https://github.com/w3c/picture-in-picture/issues/188 to track this issue. Thank you!

beaufortfrancois:
Now that https://github.com/w3c/picture-in-picture/pull/189 was merged, I think we're good to merge this PR. @marcoscaceres Does that still look good to you with my latest changes?

marcoscaceres:
Look amazing! Thank you!

PR #152, opened 2019-07-10
Remove browser icons...

marcoscaceres:
Seeing the  browser icons makes it look like everyone is implementing. That's clearly not the case.

### DIFF
--- a/implementation-status.md
+++ b/implementation-status.md
@@ -1,8 +1,6 @@
 # Implementation Status
 This document shows the current implementation status of Picture-in-Picture on the different browsers.
 
-<a href="#chrome"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/chrome/chrome_128x128.png" alt="Chrome browser logo"></a><a href="#microsoft-edge"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/edge/edge_128x128.png" alt="Microsoft Edge browser logo"></a><a href="#safari"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/safari/safari_128x128.png" alt="Safari browser logo"></a><a href="#firefox"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/firefox/firefox_128x128.png" alt="Firefox browser logo"></a>
-
 # Chrome
 
 Work is still in progress in [Chrome Canary](http://chrome.com/canary):
@@ -32,13 +30,18 @@ Tip: Chrome channel releases are tracked at [https://googlechromelabs.github.io/
 * [Issue 854935](http://crbug.com/854935): Closed captions in Picture-in-Picture are not supported.
 
 # Microsoft Edge
-https://dev.windows.com/en-us/microsoft-edge/platform/status/pictureinpicture
+Under consideration.
+https://developer.microsoft.com/en-us/microsoft-edge/platform/status/pictureinpicture/
 
 # Safari
+No public signals. 
 https://bugs.webkit.org/show_bug.cgi?id=182688
 
 # Firefox
+No public signals. 
 https://bugzilla.mozilla.org/show_bug.cgi?id=1463402
+
+Mozilla position: https://github.com/mozilla/standards-positions/issues/72
 
 # Polyfill
 https://github.com/gbentaieb/pip-polyfill/


implementation-status.md
```diff
@@ -32,13 +30,18 @@ Tip: Chrome channel releases are tracked at [https://googlechromelabs.github.io/
 * [Issue 854935](http://crbug.com/854935): Closed captions in Picture-in-Picture are not supported.
 
 # Microsoft Edge
-https://dev.windows.com/en-us/microsoft-edge/platform/status/pictureinpicture
+Under consideration.
+https://developer.microsoft.com/en-us/microsoft-edge/platform/status/pictureinpicture/
```
mounirlamouri: Edgium has the feature.

implementation-status.md
```diff
@@ -32,13 +30,18 @@ Tip: Chrome channel releases are tracked at [https://googlechromelabs.github.io/
...
 
 # Safari
+No public signals.
```
mounirlamouri: My understanding is that Safari is supportive and has their own private API.

implementation-status.md
```diff
@@ -32,13 +30,18 @@ Tip: Chrome channel releases are tracked at [https://googlechromelabs.github.io/
...
-https://dev.windows.com/en-us/microsoft-edge/platform/status/pictureinpicture
+Under consideration.
+https://developer.microsoft.com/en-us/microsoft-edge/platform/status/pictureinpicture/
 
 # Safari
+No public signals. 
 https://bugs.webkit.org/show_bug.cgi?id=182688
 
 # Firefox
+No public signals.
```
mounirlamouri: I would say there was a lot of public signals ^_^

marcoscaceres: Heh, our official(tm) position is to “defer”:
https://github.com/mozilla/standards-positions/issues/72

 That doesn’t mean we wont do it, just not right now.

implementation-status.md
marcoscaceres: Oh, do they expose some `webkitThing.` thing?

mounirlamouri: Yeah, I think it's something like `video.webkitSetPresentationMode("picture-in-picture")`.

marcoscaceres: Neat, I had no idea.

implementation-status.md
marcoscaceres: @MarionDaly, can you confirm? Edge planning to ship this too?

beaufortfrancois: Edge Canary channel has Picture-in-Picture API enabled as you can see.
image

implementation-status.md
beaufortfrancois: From https://developers.google.com/web/updates/2018/10/watch-video-using-picture-in-picture#background,

> In September 2016, Safari added Picture-in-Picture support through a WebKit API in macOS Sierra. Six months later, Chrome automatically played Picture-in-Picture video on mobile with the release of Android O using a native Android API. Six months later, we announced our intent to build and standardize a Web API, feature compatible with Safari’s, that would allow web developers to create and control the full experience around Picture-in-Picture. And here we are!

beaufortfrancois: Moreover, they're accepting patches for review: https://bugs.webkit.org/show_bug.cgi?id=189848#c8

beaufortfrancois:
I'll merge this PR for now and will start a PR for updating browsers vendor status.

implementation-status.md
marcoscaceres: Ok, nice. This is encouraging!

implementation-status.md
beaufortfrancois: FYI I've opened https://github.com/WICG/picture-in-picture/pull/153

implementation-status.md
beaufortfrancois: FYI I've opened https://github.com/WICG/picture-in-picture/pull/155

implementation-status.md
beaufortfrancois: FYI I've pushed https://github.com/WICG/picture-in-picture/pull/154

implementation-status.md
marcoscaceres: So, Mozilla is happy for you all the figure out the spec details, tests, etc. Can you please ping me when you think the spec is CR ready and it’s in Safari Tech Preview? I think then might be a good time to re-evaluate adding the API. Given that we have PiP working, adding the DOM bits should be fairly trivial (famous last words).

PR #168, opened 2019-09-12
Update README.md

cynthia:
To point to the actual URL the spec is in.


tidoust:
Marked as non substantive for IPR from ash-nazg.

beaufortfrancois:
Thank you @cynthia!

PR #174, opened 2020-01-14
Fix cross-spec links and references

tidoust:
Minor editorial updates to fix a few broken links and correct references:
- Drop custom definitions that are no longer needed
- Replace "triggered by user activation" with new user activation model (see https://github.com/whatwg/html/issues/5129)
- Use back ticks for values instead of var shorthand
- Drop link to feature name in Feature Policy (term no longer exists)

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/picture-in-picture/pull/174.html" title="Last updated on Jan 14, 2020, 4:34 PM UTC (f4b4902)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/174/4db5e80...tidoust:f4b4902.html" title="Last updated on Jan 14, 2020, 4:34 PM UTC (f4b4902)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -487,11 +487,10 @@ specification may allow PIP-ing arbitrary HTML content.
 
 ## Feature Policy ## {#feature-policy}
 
-This specification defines a <a>policy-controlled feature</a> that controls
-whether the <a>request Picture-in-Picture algorithm</a> may return a
-{{SecurityError}} and whether {{pictureInPictureEnabled}} is `true` or `false`.
-
-The feature name for this feature is `"picture-in-picture"`.
+This specification defines a <a>policy-controlled feature</a> named
+`"picture-in-picture"` that controls whether the <a>request Picture-in-Picture
+algorithm</a> may return a {{SecurityError}} and whether
+{{pictureInPictureEnabled}} is `true` or `false`.
 
 The <a>default allowlist</a> for this feature is `*`.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -501,7 +491,7 @@ This specification defines a <a>policy-controlled feature</a> that controls
 whether the <a>request Picture-in-Picture algorithm</a> may return a
 {{SecurityError}} and whether {{pictureInPictureEnabled}} is `true` or `false`.
 
-The <a>feature name</a> for this feature is `"picture-in-picture"`.
+The feature name for this feature is `"picture-in-picture"`.
```
beaufortfrancois: If "feature name" doesn't exist anymore, shall we update spec text that says "If the document is not allowed to use the policy-controlled feature named "picture-in-picture", throw a SecurityError and abort these steps." as well?

tidoust: Actually, to avoid the dubious repetition "the feature name for this feature is", I just pushed an update to merge the sentence here with the previous paragraph .

Even though the term "feature name" is no longer defined, it still seems reasonable to talk about the name of a feature, so I didn't try to update the algorithm step. If you prefer, we could perhaps simply drop "named" in both cases to end up with 'policy-controlled feature "picture-in-picture"'.

beaufortfrancois: https://github.com/w3c/picture-in-picture/pull/174/commits/f4b49022fff416c0c9f71b332db6af1affdd3766 looks good to me.

PR #177, opened 2020-02-04
Update Android's implementation status

grundmanise:
- Updated implementation status for Android. Currently PiP API is not implemented at all on android: https://bugs.chromium.org/p/chromium/issues/detail?id=806249#c63
- Added note about disabling auto PiP behaviour on Android


tidoust:
Marked as non substantive for IPR from ash-nazg.

beaufortfrancois:
Looks good. Thank you @grundmanise!

PR #186, opened 2020-06-11
Remove recommendation for minimum and maximum PiP window

guest271314:
Remove recommendation for minimum and maximum size PiP window restrictions/limitations.

There is no apparent reason for the recommendation for restricting/limiting PiP window size.

PiP window is an application to `getDisplayMedia()`. If the user wants to capture PiP window, and adjust the pixel dimensions to arbitrary sizes, they should be able to do so without arbitrary restrictions.

Note. Am **_not_** a W3C member, so their "IPR" bells and whistles will probably go off for this PR. The gist is clear. Someone with appropriate clearance can delete the language and push the change.

***
<a href="https://pr-preview.s3.amazonaws.com/guest271314/picture-in-picture/pull/186.html" title="Last updated on Jun 11, 2020, 1:55 PM UTC (fd0615b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/186/0b4bff1...guest271314:fd0615b.html" title="Last updated on Jun 11, 2020, 1:55 PM UTC (fd0615b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -211,10 +211,6 @@ Styles applied to |video| (such as opacity, visibility, transform, etc.) MUST
 NOT apply in the Picture-in-Picture window. Its aspect ratio is based on the
 video size.
 
-It is also RECOMMENDED that the Picture-in-Picture window has a maximum and
-minimum size. For example, it could be restricted to be between a quarter and
-a half of one dimension of the screen.
-
 ## Exit Picture-in-Picture ## {#exit-pip}
 
 When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked,


beaufortfrancois:
Thanks for submitting this PR @guest271314!
@mounirlamouri  Shall we remove this recommendation from the spec?

beaufortfrancois:
Without a minimum size, how can you make sure users can close the Picture-in-Picture window, or even pause video in the Picture-in-Picture window? From a security and usability perspective, I believe we should keep the minimum window size.

guest271314:
Does not `HTMLVideoElement.controls` provide that functionality?

guest271314:
Particularly at Chromium global media control UI is available. Am not sure how "security" is involved at all, can you  indicate how?

guest271314:
If maintaining a minimum PiP window size is a sticking point, keeping that recommendation would be a viable compromise that all parties should be able to move forward in agreement on. The main issue is restricting maximum PiP window size.

mounirlamouri:
FWIW, it is good practice for specifications to suggest UI best practices. A spec RECOMMENDATION is not something that *needs* to be implemented in order to be a valid implementation. This is however part of the specification because all native implementation we have seen have a minimum and maximum size and it seems reasonable to assume that these implementations, if they all agreed on a behaviour, had a good reason to do so. Chrome doesn't use a native backend on platforms other than Android but decided to follow the same pattern. I would suspect that Safari does the same.

jernoble:
>  I would suspect that Safari does the same.

That's correct.

guest271314:
> all native implementation we have seen have a minimum and maximum size

Not Firefox. 

Still have yet to read any reason for the maximum limitation.

The Chromium issue to remove the arbitrary restriction is still open.

guest271314:
> it seems reasonable to assume

Do not rely on assumptions here. Rely on evidence. So far have not seen any evidence or explanation _why_ there is a maximum restriction.

https://bugs.chromium.org/p/chromium/issues/detail?id=937859#c18

> It is still not clear at all why the restriction on PiP window size is implemented by Chromium in the first case. 

> Can the reasoning behind that decision be made clear here?

guest271314:
Am still in this PR there is no explanation as to _why_ there is a maximum width, height restriction.

Not every user is rolling around with a "smart" phone tethered to their person all day long worried about the screen being locked. 

To avoid "assume" can the maintainers of this repository at least explain _why_ they recommendation is in the specification?

guest271314:
@beaufortfrancois Can we get a straight answer to the question _**why**_ the maximum width and height restriction is in the specification and implemented at Chrome?

So far every individual involved that have asked that question has failed to provide an answer to that question, leaving the only rational conclusion that the language is 100% arbitrary, until proven otherwsie.

mounirlamouri:
@guest271314 please stop asking the same question over and over again because you do not like the answer.

This language is a spec RECOMMENDATION which is in no way a requirement but a suggestion. This suggestion comes from the experience of platforms and browsers implementing Picture-in-Picture (either the Web API or another sort). If a browser implementing this API wants to do something else, it's fully up to them.

PR #187, opened 2020-06-12
fix: check HTMLVideoElement/disablePictureInPicture attribute

### LINKED ISSUE #144: disablePictureInPicture should initialize to false
Spec says:

> 5. OPTIONALLY, if the {{disablePictureInPicture}} attribute is present on |video|, throw an {{InvalidStateError}} and abort these steps.

I think here you want to check the value of the IDL attribute value, no? It would be weird if the `disablePictureInPicture` conditionally appeared and disappeared. 

What is confusing here is that the specs seems to be talking about the HTML attribute itself being present... but what matters more (or what is easier to check) is the value being reflected in the IDL attribute (so maybe just initialize it to false).

### LINKED ISSUE #156: Integration with HTML
In #150, @domenic wrote the following. Creating an issue so it doesn't get lost. 

> Note that (if it has multi-implementer interest) the https://wicg.github.io/picture-in-picture/#htmlvideoelement-extensions section should migrate to the HTML Standard, linking to the algorithms defined in this separate specification. This would also allow the chance to fix several issues with the current specification due to insufficient integration (e.g. the use of EventHandler attributes on HTMLMediaElement instead of GlobalEventHandlers, or the way that autoPictureInPicture, disablePictureInPicture, onenterpictureinpicture, andonleavepictureinpicture have no definitions).

marcoscaceres:
closes #156 
closes #144

The check is not that "attribute is present", but that it has been set to "true"... the IDL attribute is always present, and always initializes to false. 

```
x = document.createElement("video")
x.disablePictureInPicture; // false
```

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -181,8 +181,8 @@ run the following steps:
     {{InvalidStateError}} and abort these steps.
 4. If |video| has no video track, throw a {{InvalidStateError}} and abort
     these steps.
-5. OPTIONALLY, if the {{disablePictureInPicture}} attribute is present on
-    |video|, throw an {{InvalidStateError}} and abort these steps.
+5. OPTIONALLY, if |video|'s {{HTMLVideoElement/disablePictureInPicture}} is true,
+    throw an {{InvalidStateError}} and abort these steps.
 6. If |userActivationRequired| is `true` and the <a>relevant global object</a>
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
Thank you @marcoscaceres!

Issue #188, opened 2020-06-12
leavepictureinpicture event should be fired with the Picture-in-Picture window 


beaufortfrancois:
> Ohh... um... ok... this seems like an odd API design choice here... why not fire an `EnterPictureInPictureEvent` with "leavepictureinpicture" (yes, it has "Enter" on the front, boo!) if the developer still needs to access `pictureInPictureWindow`? Otherwise, the API forces the developer to keep an unnecessary reference to a variable in the global or some outer scope?

_Originally posted by @marcoscaceres in https://github.com/w3c/picture-in-picture/pull/149_

beaufortfrancois:
This seems like a good suggestion to me. What do you think @mounirlamouri @jernoble?

> When the exit Picture-in-Picture algorithm is invoked, the user agent MUST run the following steps:
> 1. If pictureInPictureElement is null, throw a InvalidStateError and abort these steps.
> 1. Run the close window algorithm with the Picture-in-Picture window associated with pictureInPictureElement.
> 1. Unset pictureInPictureElement.
> 1. ~~Queue a task to fire an event with the name `leavepictureinpicture` at the video with its bubbles attribute initialized to true.~~
> 1. __Queue a task to fire an event with the name `leavepictureinpicture` using `PictureInPictureEvent` at the video with its bubbles attribute initialized to true and its `pictureInPictureWindow` attribute initialized to "Picture-in-Picture window".__

beaufortfrancois:
For info, here's the potential spec change: https://github.com/w3c/picture-in-picture/pull/189

mounirlamouri:
@jernoble any concern from Apple here?

jernoble:
LGTM.

PR #189, opened 2020-06-12
leavepictureinpicture event fires with Picture-in-Picture window

beaufortfrancois:
FIX https://github.com/w3c/picture-in-picture/issues/188


beaufortfrancois:
Web platform tests changes are also available in https://chromium-review.googlesource.com/c/chromium/src/+/2241537

foolip:
I have filed https://bugs.webkit.org/show_bug.cgi?id=221083 about renaming `EnterPictureInPictureEvent` to `PictureInPictureEvent`. It turns out this shipped in Safari 13.1, and probably nobody noticed the rename in the spec.

When making changes like this, it would be great to file issues for all engines that are affected. (`Source/WebCore/Modules/pictureinpicture/EnterPictureInPictureEvent.idl` was in WebKit at the time this issue was filed.)

PR #190, opened 2020-06-17
Provide a clear definition of autoPictureInPictureElement

### LINKED ISSUE #115: Provide a clear definition of autoPictureInPictureElement
In https://github.com/WICG/picture-in-picture/pull/112/files#r249006604, it was agreed to provide a clear definiton of `autoPictureInPictureElement`.

beaufortfrancois:
@mounirlamouri PTAL

Fix #115

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -279,9 +279,19 @@ added to the list of content attributes for video elements.
 The {{autoPictureInPicture}} IDL attribute MUST <a>reflect</a> the content
 attribute of the same name.
 
-The <dfn>autoPictureInPictureElement</dfn> is the video element, among all
-video elements with the {{autoPictureInPicture}} attribute currently set,
-whose {{autoPictureInPicture}} attribute was set most recently. Note that the
+All <a>documents</a> have an associated <dfn>list of Auto Picture-in-Picture
+elements</dfn>, which is an <a>ordered set</a> of {{HTMLVideoElement}} items. It
+is initially empty.
+When the {{autoPictureInPicture}} attribute is added to a video element, the
+video element is first <a>removed</a> then <a>appended</a> into the
+appropriate <a>list of Auto Picture-in-Picture elements</a>.
+When a video element is removed from the {{document}} or its
+{{autoPictureInPicture}} attribute is removed, the video element is
+<a>removed</a> from the appropriate <a>list of Auto Picture-in-Picture
+elements</a>.
+
+The <dfn>autoPictureInPictureElement</dfn> is the last video element in the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -279,9 +279,19 @@ added to the list of content attributes for video elements.
...
-The <dfn>autoPictureInPictureElement</dfn> is the video element, among all
-video elements with the {{autoPictureInPicture}} attribute currently set,
-whose {{autoPictureInPicture}} attribute was set most recently. Note that the
+All <a>documents</a> have an associated <dfn>list of Auto Picture-in-Picture
+elements</dfn>, which is an <a>ordered set</a> of {{HTMLVideoElement}} items. It
+is initially empty.
+When the {{autoPictureInPicture}} attribute is added to a video element, the
+video element is first <a>removed</a> then <a>appended</a> into the
+appropriate <a>list of Auto Picture-in-Picture elements</a>.
+When a video element is removed from the {{document}} or its
```
mounirlamouri, line 288: Removing from the document removes the element from the list but not appending to the document. Why?

beaufortfrancois: I assumed "remove from the document" meant garbage-collected.

mounirlamouri: No, I meant `element.parentElement.removeChild(element);`.

beaufortfrancois: As long as the element is still alive, I'm not sure we'd want to remove it from this list even with `element.parentElement.removeChild(element);`.

index.bs
```diff
@@ -279,9 +279,19 @@ added to the list of content attributes for video elements.
 The {{autoPictureInPicture}} IDL attribute MUST <a>reflect</a> the content
 attribute of the same name.
 
-The <dfn>autoPictureInPictureElement</dfn> is the video element, among all
-video elements with the {{autoPictureInPicture}} attribute currently set,
-whose {{autoPictureInPicture}} attribute was set most recently. Note that the
+All <a>documents</a> have an associated <dfn>list of Auto Picture-in-Picture
```
chrisn, line 282: It's not clear whether this means there's a single list per-document, or one list across all documents. Also, it's not clear what *appropriate* means here, is this referring to a per-document list?

index.bs
mounirlamouri: The only way to do that would be to have the list holds a weak reference to the element which I wouldn't know how to express in spec terms. Otherwise, being in that list would prevent the element from being GC'd.

beaufortfrancois:
https://github.com/w3c/picture-in-picture/pull/216 makes this PR obsolete.

Issue #217, opened 2022-12-15
Page Visibility spec has been merged into HTML


rakuco:
https://w3c.github.io/picture-in-picture/#page-visibility mentions the Page Visibility specification. This specification was folded into the main HTML spec in 2021 or so. The references there (including the reference to `visibilityState`) need to be updated.

beaufortfrancois:
Thanks again for raising this issue @rakuco!

PR #218, opened 2022-12-19
Remove Page Visibility spec references

beaufortfrancois:
@rakuco Thanks for filing https://github.com/w3c/picture-in-picture/issues/217

Hopefully this PR fixes it. Can you have a look?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/pull/218.html" title="Last updated on Dec 19, 2022, 2:51 PM UTC (b9df8a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/218/a4e812a...b9df8a2.html" title="Last updated on Dec 19, 2022, 2:51 PM UTC (b9df8a2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -25,7 +25,6 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: media element event task source
     type: dfn
         urlPrefix: interaction.html
-            text: traversable navigable
             text: system visibility state
 spec: Fullscreen; urlPrefix: https://fullscreen.spec.whatwg.org
     type: dfn; text: fullscreen flag; url: #fullscreen-flag
@@ -314,7 +313,7 @@ window.
 
 The Picture-in-Picture window visibility MUST NOT be taken into account by the
 user agent to determine if the <a>system visibility state</a> of a 
-<a>traversable navigable</a> has changed.
+<a for="/">traversable navigable</a> has changed.
 
 ## One Picture-in-Picture window ## {#one-pip-window}
 


rakuco (commented):
Looking at https://html.spec.whatwg.org/multipage/interaction.html#page-visibility, `{{visibilityState}}` just returns the value of an internal slot that's set to either "hidden" or "visible".

Wouldn't it make more sense to say that "the PiP window visibility must not be taken into account by the user agent to determine if the system visibility state of a traversable navigable has changed" or something along those lines?

beaufortfrancois:
@rakuco Feedback addressed! Thank you :)

index.bs
```diff
@@ -23,6 +23,10 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn
         urlPrefix: media.html
             text: media element event task source
+    type: dfn
+        urlPrefix: interaction.html
+            text: traversable navigable
```
rakuco, line 28: This <dfn> is exported by the HTML, so I don't think you need to add it here explicitly.

rakuco (approved):
lgtm with a nit. Thanks!

index.bs
beaufortfrancois: I had to because of:

```
LINK ERROR: No 'dfn' refs found for 'system visibility state' that are marked for export.
<a data-link-type="dfn" data-lt="system visibility state">system visibility state</a>
LINK ERROR: Multiple possible 'traversable navigable' dfn refs.
Arbitrarily chose https://html.spec.whatwg.org/multipage/document-sequences.html#nav-traversable
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:html; type:dfn; for:navigable; text:traversable navigable
spec:html; type:dfn; for:/; text:traversable navigable
<a data-link-type="dfn" data-lt="traversable navigable">traversable navigable</a>
LINK ERROR: No 'dfn' refs found for 'context object' that are marked for export.
```

rakuco: "system visibility state" isn't exported, but "traversable navigable" is. The error you're getting for the latter can probably be solved by using `<a for=/>traversable navigable</a>` like https://github.com/whatwg/storage/blob/main/storage.bs#L259 does.

beaufortfrancois: Fixed. Thanks!

rakuco (approved):
lgtm :shipit:

Issue #191, opened 2020-08-13
Remove monkey patch sections


npm1:
There are 3 sections that monkey patch other specifications (the ones with name starting with "Interaction with "). The affected specs should probably be updated instead, and those sections should be removed.

chrisn:
Thanks for raising this, and apologies for taking so long to respond. I'm noting the [TAG guidance](https://w3ctag.github.io/design-principles/#monkey-patching), so looking at each of those sections:

* Interaction with Fullscreen - It could well make sense to add to the Fullscreen spec.

* Interaction with Remote Playback - This seems informative, so could be added to the Note after [local playback device](https://w3c.github.io/remote-playback/#dfn-local-playback-device), e.g:
  > NOTE: A local playback device might have extra outputs, like an external display or speakers/headphones, or a [Picture-in-Picture window](https://www.w3.org/TR/picture-in-picture/).

* Interaction with Media Session - I'm not sure what this section is saying. The only thing I see in Media Session is the change proposed in https://github.com/w3c/mediasession/pull/295. Do we need this section at all?

* Interaction with Page Visibility - The first paragraph should probably remain in this spec as it's about the PiP window itself. The second paragraph, on system visibility state of a traversable navigable, could be added to the [Page Visibility section](https://html.spec.whatwg.org/multipage/interaction.html#page-visibility) in HTML.

@beaufortfrancois, @jernoble, what do you think?

chrisn:
Discussed "Interaction with Remote Playback" in the Second Screen WG meeting at TPAC. The group agreed to move this text to the Remote Playback API spec, but will need discussion if it needs to be normative.

mfoltzgoogle:
To clarify: Second Screen expects someone working on the PiP spec to open a PR against the Remote Playback API.

chrisn:
Regarding **Interaction with Page Visibility**, #223 partially addresses this, but would leave this text remaining:

> The Picture-in-Picture window visibility MUST NOT be taken into account by the user agent to determine if the [system visibility state](https://html.spec.whatwg.org/multipage/interaction.html#system-visibility-state) of a [traversable navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#traversable-navigable) has changed.

@beaufortfrancois and @marcoscaceres Any thoughts on what to do with this? Should it remain somewhere in this spec, or could be added to the [Page Visibility section](https://html.spec.whatwg.org/multipage/interaction.html#page-visibility) in HTML?

[...]

PR #223, opened 2024-03-21
Move requirements for PiP window visibility and closing

chrisn:
This PR is in response to #191 and the Interaction with Page Visibility section. It moves the part about the PiP window being visible to the Request Picture-in-Picture section, as well as the recommendation to provide a way for users to close the PiP window.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -209,14 +209,16 @@ the user agent MUST run the following steps:
 8. Set {{pictureInPictureElement}} to |video|.
 9. Let <dfn>Picture-in-Picture window</dfn> be a new instance of
     {{PictureInPictureWindow}} associated with {{pictureInPictureElement}}.
-10. Append <a>relevant settings object</a>'s <a>origin</a> to
+10. When the Picture-in-Picture window is opened it MUST be visible, even when
+    the <a>Document</a> is not in focus or hidden.
+11. Append <a>relevant settings object</a>'s <a>origin</a> to
     <a>initiators of active Picture-in-Picture sessions</a>.
-11. <a>Queue a task</a> to <a>fire an event</a> named
+12. <a>Queue a task</a> to <a>fire an event</a> named
     {{enterpictureinpicture}} using {{PictureInPictureEvent}} at the
     |video| with its {{bubbles}} attribute initialized to `true` and its
     {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
     <a>Picture-in-Picture window</a>.
-12. If {{pictureInPictureElement}} is <a>fullscreenElement</a>, it is
+13. If {{pictureInPictureElement}} is <a>fullscreenElement</a>, it is
     RECOMMENDED to <a>exit fullscreen</a>.
 
 It is RECOMMENDED that video frames are not rendered in the page and in the
@@ -236,6 +238,9 @@ video size.
 It is also RECOMMENDED that the Picture-in-Picture window has a maximum and
 minimum size. For example, it could be restricted to be between a quarter and
 a half of one dimension of the screen.
+
+The user agent SHOULD provide a way for users to manually close the
+Picture-in-Picture window.
 
 ## Exit Picture-in-Picture ## {#exit-pip}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -209,14 +209,16 @@ the user agent MUST run the following steps:
 8. Set {{pictureInPictureElement}} to |video|.
 9. Let <dfn>Picture-in-Picture window</dfn> be a new instance of
     {{PictureInPictureWindow}} associated with {{pictureInPictureElement}}.
-10. Append <a>relevant settings object</a>'s <a>origin</a> to
+10. When the Picture-in-Picture window is opened it MUST be visible, even when
```
beaufortfrancois, line 212: I'm not sure this is the right place actually.

beaufortfrancois: How about simply moving the following at the end of `## Request Picture-in-Picture ## {#request-pip}` section?

```
When {{pictureInPictureElement}} is set, the Picture-in-Picture window MUST
be visible, even when the <a>Document</a> is not in focus or hidden. The user
agent SHOULD provide a way for users to manually close the Picture-in-Picture
window.
```

chrisn: Yes, that would work. I was in two minds about it when I made this edit...

chrisn: I updated it as you suggested.

beaufortfrancois (approved):
LGTM. Thank you!

beaufortfrancois:
@chrisn I'll let you merge when you're ready.

chrisn:
Perfect, thanks. Just waiting on confirmation from @marcoscaceres.

index.bs
```diff
@@ -237,6 +237,11 @@ It is also RECOMMENDED that the Picture-in-Picture window has a maximum and
 minimum size. For example, it could be restricted to be between a quarter and
 a half of one dimension of the screen.
 
+When {{pictureInPictureElement}} is set, the Picture-in-Picture window MUST
+be visible, even when the <a>Document</a> is not in focus or hidden. The user
+agent SHOULD provide a way for users to manually close the Picture-in-Picture
+window.
```
marcoscaceres, line 243: I don't think this is quite right... or not stringent enough. I think what needs to happen is that:

* The pip window needs to be classified as a [=browsing context/top-level traversable=]. 
* The user agent MUST retain a  system visibility state if "visible" for the pip window - and should be above all other windows.

chrisn: Reading the [relevant parts of the HTML spec](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-and-session-history) I'm not sure if a PiP window should be a [top-level traversable](https://html.spec.whatwg.org/multipage/document-sequences.html#top-level-traversables), which makes it a [traversable navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#traversable-navigable), hence a [navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#navigables) which "presents a Document to the user via its active session history entry". The reason I say this is that a PiP window doesn't present a Document (it's just a rendering surface for video) and doesn't affect session history AFAIK.

index.bs
```diff
@@ -237,6 +237,11 @@ It is also RECOMMENDED that the Picture-in-Picture window has a maximum and
 minimum size. For example, it could be restricted to be between a quarter and
 a half of one dimension of the screen.
 
+When {{pictureInPictureElement}} is set, the Picture-in-Picture window MUST
+be visible, even when the <a>Document</a> is not in focus or hidden. The user
```
marcoscaceres, line 241:
```suggestion
be visible, even when the [=document=]'s {{Window}} is not in [=Window/focus=] or the  [=document=]'s [=Document/visibility state=] is "hidden". The user
```

index.bs
marcoscaceres: Right, yes... it's not a navigable, but I think I think this should maybe run:

> Run the [focusing steps](https://html.spec.whatwg.org/#focusing-steps) with the Picture-in-Picture
window.

To focus it. It's still a Window and Document with some limited functionality.

chrisn: I added the suggested change, but the 'focused' link doesn't look right yet.

Issue #226, opened 2024-06-07
Typo: Return promise vs resolve promise


joeyparrish:
I believe here: https://github.com/w3c/picture-in-picture/blob/8e590950301a7379d7de1ec64c50324526280c85/index.bs#L353

> _Return_ promise ...

Should say:

> _Resolve_ promise ...

marcoscaceres:
Agree. Further, retuning the promise should be step 1, then go in parallel, then eventually queue a task to resolve it.

beaufortfrancois:
Thanks @joeyparrish for catching this!
I've opened https://github.com/w3c/picture-in-picture/pull/227 to address this issue.

@marcoscaceres The first sentence says `The {{requestPictureInPicture()}} method, when invoked, MUST return <a>a new promise</a> |promise| and run the following steps <a>in parallel</a>:` which I believe should be enough.

PR #227, opened 2024-06-10
Replace return with resolve

beaufortfrancois:
FIX https://github.com/w3c/picture-in-picture/issues/226

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/pull/227.html" title="Last updated on Jun 10, 2024, 6:58 AM UTC (b8bcff0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/227/8e59095...b8bcff0.html" title="Last updated on Jun 10, 2024, 6:58 AM UTC (b8bcff0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -350,7 +350,7 @@ parallel</a>:
 2. Run the <a>request Picture-in-Picture algorithm</a> with |video|.
 3. If the previous step threw an exception, reject |promise| with that
     exception and abort these steps.
-4. Return |promise| with the <a>Picture-in-Picture window</a> associated with
+4. [=/Resolve=] |promise| with the <a>Picture-in-Picture window</a> associated with
     {{pictureInPictureElement}}.
 
 ## Extensions to <code>Document</code> ## {#document-extensions}
@@ -378,7 +378,7 @@ parallel</a>:
 1. Run the <a>exit Picture-in-Picture algorithm</a>.
 2. If the previous step threw an exception, reject |promise| with that
     exception and abort these steps.
-3. Return |promise|.
+3. [=/Resolve=] |promise|.
 
 ## Extension to <code>DocumentOrShadowRoot</code> ## {#documentorshadowroot-extension}
 


padenot (approved):
Obvious fix, looks good.

chrisn (approved):

index.bs
```diff
@@ -378,7 +378,7 @@ parallel</a>:
 1. Run the <a>exit Picture-in-Picture algorithm</a>.
 2. If the previous step threw an exception, reject |promise| with that
     exception and abort these steps.
```
marcoscaceres, line 380:
```suggestion
2. If the previous step [=exception/throws|threw=] an exception, [=queue a task=] using [=this=]'s [=media element event task source=] to [=/reject=] |promise| with that exception and abort these steps.
```

index.bs
```diff
@@ -350,7 +350,7 @@ parallel</a>:
 2. Run the <a>request Picture-in-Picture algorithm</a> with |video|.
 3. If the previous step threw an exception, reject |promise| with that
     exception and abort these steps.
-4. Return |promise| with the <a>Picture-in-Picture window</a> associated with
+4. [=/Resolve=] |promise| with the <a>Picture-in-Picture window</a> associated with
     {{pictureInPictureElement}}.
```
marcoscaceres, line 354:
```suggestion
4. [=Queue a task=] using |video|'s [=media element event task source=] to  [=/resolve=] |promise| with the <a>Picture-in-Picture window</a> associated with |video|'s {{pictureInPictureElement}}.
```

index.bs
```diff
@@ -350,7 +350,7 @@ parallel</a>:
 2. Run the <a>request Picture-in-Picture algorithm</a> with |video|.
 3. If the previous step threw an exception, reject |promise| with that
     exception and abort these steps.
```
marcoscaceres, line 352:
```suggestion
3. If the previous step [=exception/throws|threw=] an exception, [=queue a task=] using |video|'s [=media element event task source=] to [=/reject=] |promise| with that exception and abort these steps.
```

marcoscaceres (commented):
Should be like this... also, we should make better use of [=this=], but we can do that as a followup. I need to go through the document and fix up how internal things are accessed.

index.bs
beaufortfrancois: Both WebKit and Chrome do not queue a task from what I can see at https://github.com/WebKit/WebKit/blob/d36ed108e92fb350723da8e7a88e7a7e8be1ef45/Source/WebCore/Modules/pictureinpicture/DocumentPictureInPicture.cpp#L54 and https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/picture_in_picture/document_video_picture_in_picture.cc;l=43;drc=95cf77bcd2a1f44e564e8404d2d363f508d580ce

I'm not sure we want to update both implementations for this.

index.bs
beaufortfrancois: For both resolve and reject paths, WebKit and Chrome don't queue task. See https://github.com/WebKit/WebKit/blob/d36ed108e92fb350723da8e7a88e7a7e8be1ef45/Source/WebCore/Modules/pictureinpicture/HTMLVideoElementPictureInPicture.cpp#L81 and https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/document_picture_in_picture/picture_in_picture_controller_impl.cc;l=124;drc=20135c10f0869fdefb75d990ec84143a649d84c3

index.bs
marcoscaceres: Wait! Then this algorithm is wrong then. Above it says it runs “in parallel”, and you can see from both implementations that it’s running on the main thread. 

This needs be updated to return a promise rejected with X and the later run the steps in parallel.

beaufortfrancois: The reject path runs on the main thread because it's easy to catch some errors but the resolve path and some reject paths are not.

index.bs
marcoscaceres: This should be:
```
The {{requestPictureInPicture()}} method, when invoked, MUST run the following steps:

1. Run the <a>request Picture-in-Picture algorithm</a> with [=this=].
2. If the previous step [=exception/throws|threw=] an exception, return [=a promise rejected with=] that exception.
3. return [=a promise resolved with=] [=this=]'s associated [=Picture-in-Picture window=].
```

As it never seems to go into parallel.

index.bs
marcoscaceres: We need to split those up then. This should match implementations. I can see like 6 reject cases that run on the main thread. Only see this in WebKit getting sent off main thread: 

```C++ 
    if (videoElement.webkitSupportsPresentationMode(HTMLVideoElement::VideoPresentationMode::PictureInPicture)) {
        videoElementPictureInPicture->m_enterPictureInPicturePromise = WTFMove(promise);
        videoElement.webkitSetPresentationMode(HTMLVideoElement::VideoPresentationMode::PictureInPicture);
    }
```

marcoscaceres: Even there... we would need to check what videoElement.webkitSetPresentationMode does and if it goes off main thread.

joeyparrish:
Thanks @beaufortfrancois for addressing this!

beaufortfrancois:
What shall we do about this PR?
It seems like updating the spec with suggested changes from @marcoscaceres would make current implementations not compliant.

beaufortfrancois:
If that's okay with everyone, I'll merge this PR and will open an issue about @marcoscaceres's feedback

joeyparrish (approved):

beaufortfrancois:
I've opened https://github.com/w3c/picture-in-picture/issues/229

PR #233, opened 2024-09-26
Fix tasks for enterPictureInPicture and exitPictureInPicture.

### LINKED ISSUE #225: Disable Picture-in-Picture steps should queue a task
Raised by @marcoscaceres in #224, about the [Disable Picture-in-Picture](https://w3c.github.io/picture-in-picture/#disable-pip) steps:

> This should queue a task or something... promises can't be rejected synchronously here.
> 
> Consider, what should happen here:
>
> const video = document.querySelector("video"); 
> await video.requestPictureInPicture();
> video.setAttribute("disablePictureInPicture", "");
> video.removeAttribute("disablePictureInPicture");
> 
> That should probably have no effect. The task at the turn of the event loop should check if the attribute is still present and reject the promise(s) at that point.

### LINKED ISSUE #229: Investigate if the algorithm and the implementations need to be updated
As originally posted by @marcoscaceres in https://github.com/w3c/picture-in-picture/pull/227#discussion_r1635906594:

> this algorithm is wrong then. Above it says it runs “in parallel”, and you can see from both implementations that it’s running on the main thread. 
>This needs be updated to return a promise rejected with X and the later run the steps in parallel.

We need to investigate if the algorithm and the implementations need to be updated.

### LINKED ISSUE #230: Missing tasks in parallel steps in Picture-in-Picture
While crawling [Picture-in-Picture](https://w3c.github.io/picture-in-picture/), the following algorithms fire an event, or resolve or reject a Promise, within a step that runs [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) without first queuing a task:
* [ ] The algorithm that starts with "The requestPictureInPicture() method, when invoked, MUST return a new promise promise and run the following steps in parallel:" resolves/rejects a promise directly in a step that runs in parallel
* [ ] The algorithm that starts with "The exitPictureInPicture() method, when invoked, MUST return a new promise promise and run the following steps in parallel:" resolves/rejects a promise directly in a step that runs in parallel

See [Dealing with the event loop](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors) in the HTML specification for guidance on how to deal with algorithm sections that run *in parallel*.

<sub>Cc @dontcallmedom @tidoust</sub>

<sub>This issue was detected and reported semi-automatically by [Strudy](https://github.com/w3c/strudy/) based on data collected in [webref](https://github.com/w3c/webref/).</sub>

markafoltz:
Closes #230: ensures that Promises returned by these methods use the media element task source.

Also fixes several ambiguous cross references.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/pull/233.html" title="Last updated on Sep 26, 2024, 11:57 PM UTC (5d9f665)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/233/e819c96...5d9f665.html" title="Last updated on Sep 26, 2024, 11:57 PM UTC (5d9f665)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -34,14 +34,21 @@ spec: Remote-Playback; urlPrefix: https://w3c.github.io/remote-playback/#dfn-
     type: dfn
         text: local playback device
         text: local playback state
+spec: ecmascript; urlPrefix: https://tc39.es/ecma262/multipage/
+    type: dfn;
+        text: current realm; url: executable-code-and-execution-contexts.html#sec-code-realms
 </pre>
 
 <pre class="link-defaults">
 spec:dom; type:attribute; text:bubbles
 spec:dom; type:dfn; for:NamedNodeMap; text:element
 spec:dom; type:interface; text:Document
+spec:dom; type:dfn; text:origin
 spec:html; type:attribute; for:HTMLMediaElement; text:readyState
 spec:html; type:dfn; for:/; text:browsing context
+spec:html; type:dfn; for:realm; text:global object
+spec:infra; type:dfn; text:user agent
+spec:infra; type:dfn; for:list; text:item
 </pre>
 
 # Introduction # {#intro}
@@ -207,7 +214,7 @@ the user agent MUST run the following steps:
     {{PictureInPictureWindow}} associated with {{pictureInPictureElement}}.
 10. Append <a>relevant settings object</a>'s <a>origin</a> to
     <a>initiators of active Picture-in-Picture sessions</a>.
-11. <a>Queue a task</a> to <a>fire an event</a> named
+11. <a>Queue a picture-in-picture task</a> to <a>fire an event</a> named
     {{enterpictureinpicture}} using {{PictureInPictureEvent}} at the
     |video| with its {{bubbles}} attribute initialized to `true` and its
     {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
@@ -242,7 +249,7 @@ the user agent MUST run the following steps:
     abort these steps.
 2. Run the <a>close window algorithm</a> with the <a>Picture-in-Picture
     window</a> associated with {{pictureInPictureElement}}.
-3. <a>Queue a task</a> to <a>fire an event</a> named
+3. <a>Queue a picture-in-picture task</a> to <a>fire an event</a> named
     {{leavepictureinpicture}} using {{PictureInPictureEvent}} at the
     |video| with its {{bubbles}} attribute initialized to `true` and its
     {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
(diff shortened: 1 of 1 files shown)


markafoltz:
Should also close #225

beaufortfrancois:
@markafoltz From what I can see in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/document_picture_in_picture/picture_in_picture_controller_impl.cc;l=158?q=kMediaElementEvent%20f:picture&ss=chromium%2Fchromium%2Fsrc, Chromium's implementation will no longer comply with the spec. I may be wrong about this but is this what we want ?

Note that this PR is related to https://github.com/w3c/picture-in-picture/issues/229

markafoltz:
It should be impossible to fire events or resolve promises off of the main thread in Blink, so where do you see Chrome not complying?  (Sorry I don't know the PiP code.)

markafoltz:
Thanks for the pointer as well, this should also close #229

beaufortfrancois:
> It should be impossible to fire events or resolve promises off of the main thread in Blink, so where do you see Chrome not complying? (Sorry I don't know the PiP code.)

Like I said, I may be wrong, but https://source.chromium.org/search?q=kMediaElementEvent%20f:picture&sq=&ss=chromium%2Fchromium%2Fsrc returns results where we specifically use the media element task source in Picture-in-Picture, and I'm not sure it covers all the cases you've updated in this PR.

An example would be `the user agent MUST <a>queue a picture-in-picture task</a> to <a>fire an event</a> named {{resize}} at {{pictureInPictureElement}}.` at https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/picture_in_picture/picture_in_picture_window.cc;l=30;drc=f522344e45882da4c7f7cb1b3a0a7bd747d654bb
I believe we should be using something like `EnqueueEvent(*Event::Create(event_type_names::kResize), TaskType::kMediaElementEvent);` instead of synchronously dispatching the event with `DispatchEvent(*Event::Create(event_type_names::kResize));` like we do today.

index.bs
```diff
@@ -455,6 +464,11 @@ dictionary PictureInPictureEventInit : EventInit {
 The <a>task source</a> for all the tasks queued in this specification is the
 <a>media element event task source</a> of the video element in question.
 
+When an algorithm <dfn lt="queue a picture-in-picture task">queues a
```
marcoscaceres: I wonder if we should just reuse [[queue a media element task]] instead: 
https://html.spec.whatwg.org/multipage/media.html#queue-a-media-element-task

That takes a media element instead it seems... that might make more sense?

index.bs
```diff
@@ -376,9 +383,10 @@ return <a>a new promise</a> |promise| and run the following steps <a>in
 parallel</a>:
 
 1. Run the <a>exit Picture-in-Picture algorithm</a>.
-2. If the previous step threw an exception, reject |promise| with that
+2. If the previous step threw an exception,
```
```suggestion
2. If the previous step threw an [=exception=],
```

marcoscaceres:
Ok, I'm going to say #245 closes #233.

PR #241, opened 2025-09-18
Expand permissions policy to include auto-pip

bpkeen:
This PR modifies the **Permissions Policy** section to include a new policy for automatic picture-in-picture. The policy will allow controlling whether the user agent can automatically enter picture-in-picture.

***
<a href="https://pr-preview.s3.amazonaws.com/bpkeen/picture-in-picture/pull/241.html" title="Last updated on Jul 29, 2026, 12:42 AM UTC (e961626)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/241/46651d5...bpkeen:e961626.html" title="Last updated on Jul 29, 2026, 12:42 AM UTC (e961626)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -493,7 +493,13 @@ whether picture-in-picture is allowed:
 
 <dl dfn-type=value>
   <dt>"<dfn permission export>picture-in-picture</dfn>".
+  <dd>
+    The `"picture-in-picture"` permission policy.
+
   <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
+
+  <dd>
+    The `"automatic-picture-in-picture"` permission policy.
 </dl>
 
 Access to the Picture-in-Picture API is gated behind the
--- a/index.bs
+++ b/index.bs
@@ -34,10 +34,6 @@ spec: Remote-Playback; urlPrefix: https://w3c.github.io/remote-playback/#dfn-
     type: dfn
         text: local playback device
         text: local playback state
-spec: MediaSession; urlPrefix: https://w3c.github.io/mediasession
-    type: dfn; text: enterpictureinpicture; url: #dom-mediasessionaction-enterpictureinpicture
-    type: dfn; text: contentoccluded; url: #dom-mediasessionenterpictureinpicturereason-contentoccluded
-    type: dfn; text: media session action; url: #media-session-action
 </pre>
 
 <pre class="link-defaults">
@@ -308,10 +304,10 @@ Picture-in-Picture.
 The API will have to be used with the [[MediaSession]] API for customizing the
 available controls on the Picture-in-Picture window.
 
-The user agent SHOULD not override the <a>media session action</a>
-<a>enterpictureinpicture</a> action handler,  if one is provided by the site.
-This gives the site control over the Picture-in-Picture user experience, when a
-Picture-in-Picture window is requested.
+The user agent SHOULD not override the media session action
+{{MediaSessionAction/"enterpictureinpicture"}} action handler,  if one is
+provided by the site. This gives the site control over the Picture-in-Picture
+user experience, when a Picture-in-Picture window is requested.
 
 ## Interaction with Page Visibility ## {#page-visibility}
 
@@ -497,22 +493,28 @@ whether picture-in-picture is allowed:
 <dl dfn-type=value>
   <dt>"<dfn permission export>picture-in-picture</dfn>".
   <dd>
-    The `"picture-in-picture"` permission policy.
+    Controls whether the document is allowed to use Picture-in-Picture.
+    Specifically, the <a>request Picture-in-Picture algorithm</a> throws a
+    {{SecurityError}} if the document is not [=allowed to use=] this feature.
+    The <a>default allowlist</a> for this feature is `*`.
 
   <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
   <dd>
-    The `"automatic-picture-in-picture"` permission policy.
+    Controls whether the user agent SHOULD automatically trigger
+    Picture-in-Picture. The <a>default allowlist</a> for this feature is
+    `self`.
 </dl>
 
 Access to the Picture-in-Picture API is gated behind the
 <a>policy-controlled feature</a> {{picture-in-picture}}, which has a
 <a>default allowlist</a> of `*`.
 
-When the document is not <a>allowed to use</a> the
+When the document is not [=allowed to use=] the
 <a>policy-controlled feature</a> named {{automatic-picture-in-picture}}, the
-user agent SHOULD not trigger a [[MediaSession]] <a>enterpictureinpicture</a>
-action with the <a>contentoccluded</a> reason. The <a>default allowlist</a> for
-this feature is `self`.
+user agent SHOULD not trigger a [[MediaSession]]
+{{MediaSessionAction/"enterpictureinpicture"}} action with the
+{{MediaSessionEnterPictureInPictureReason/"contentoccluded"}} reason. The
+<a>default allowlist</a> for this feature is `self`.
 
 Note: User agents should surface a console warning whenever
 `"picture-in-picture"` is blocked by the {{automatic-picture-in-picture}}
--- a/index.bs
+++ b/index.bs
@@ -305,7 +305,7 @@ The API will have to be used with the [[MediaSession]] API for customizing the
 available controls on the Picture-in-Picture window.
 
 The user agent SHOULD not override the media session action
-{{MediaSessionAction/"enterpictureinpicture"}} action handler,  if one is
+{{MediaSessionAction/"enterpictureinpicture"}} action handler, if one is
 provided by the site. This gives the site control over the Picture-in-Picture
 user experience, when a Picture-in-Picture window is requested.
 
@@ -493,10 +493,8 @@ whether picture-in-picture is allowed:
 <dl dfn-type=value>
   <dt>"<dfn permission export>picture-in-picture</dfn>".
   <dd>
-    Controls whether the document is allowed to use Picture-in-Picture.
-    Specifically, the <a>request Picture-in-Picture algorithm</a> throws a
-    {{SecurityError}} if the document is not [=allowed to use=] this feature.
-    The <a>default allowlist</a> for this feature is `*`.
+    Controls whether the [=document=] [=allowed to use|is allowed=] to use
+    Picture-in-Picture. The <a>default allowlist</a> for this feature is `*`.
 
   <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
   <dd>
--- a/index.bs
+++ b/index.bs
@@ -46,6 +46,7 @@ spec:dom; type:dfn; for:NamedNodeMap; text:element
 spec:dom; type:dfn; text:origin
 spec:dom; type:interface; text:Document
 spec:html; type:attribute; for:HTMLMediaElement; text:readyState
+spec:html; type:dfn; text:allowed to use
 spec:html; type:dfn; for:/; text:browsing context
 spec:infra; type:dfn; text:user agent
 spec:infra; type:dfn; for:list; text:item
@@ -197,8 +198,8 @@ the user agent MUST run the following steps:
 1. If <a>Picture-in-Picture support</a> is `false`, throw a
     {{NotSupportedError}} and abort these steps.
 2. If the document is not <a>allowed to use</a> the <a>policy-controlled feature</a>
-    named {{picture-in-picture}} or {{automatic-picture-in-picture}}, throw a
-    {{SecurityError}} and abort these steps.
+    named `"picture-in-picture"`, throw a {{SecurityError}} and abort these
+    steps.
 3. If |video|'s {{readyState}} attribute is {{HAVE_NOTHING}}, throw a
     {{InvalidStateError}} and abort these steps.
 4. If |video| has no video track, throw a {{InvalidStateError}} and abort
@@ -309,6 +310,8 @@ available controls on the Picture-in-Picture window.
 
 The user agent SHOULD not override the <a>media session action</a>
 <a>enterpictureinpicture</a> action handler,  if one is provided by the site.
+This gives the site control over the Picture-in-Picture user experience, when a
+Picture-in-Picture window is requested.
 
 ## Interaction with Page Visibility ## {#page-visibility}
 
@@ -497,7 +500,6 @@ whether picture-in-picture is allowed:
     The `"picture-in-picture"` permission policy.
 
   <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
-
   <dd>
     The `"automatic-picture-in-picture"` permission policy.
 </dl>
@@ -506,10 +508,11 @@ Access to the Picture-in-Picture API is gated behind the
 <a>policy-controlled feature</a> {{picture-in-picture}}, which has a
 <a>default allowlist</a> of `*`.
 
-When the {{automatic-picture-in-picture}} <a>policy-controlled feature</a>
-policy is set, the user agent SHOULD not trigger a [[MediaSession]]
-<a>enterpictureinpicture</a> action with the <a>contentoccluded</a> reason. The
-<a>default allowlist</a> for this feature is `self`.
+When the document is not <a>allowed to use</a> the
+<a>policy-controlled feature</a> named {{automatic-picture-in-picture}}, the
+user agent SHOULD not trigger a [[MediaSession]] <a>enterpictureinpicture</a>
+action with the <a>contentoccluded</a> reason. The <a>default allowlist</a> for
+this feature is `self`.
 
 Note: User agents should surface a console warning whenever
 `"picture-in-picture"` is blocked by the {{automatic-picture-in-picture}}
@@ -518,5 +521,5 @@ permissions policy.
 # Acknowledgments # {#acknowledgments}
 
 Thanks to Jennifer Apacible, Zouhir Chahoud, Marcos Cáceres, Philip Jägenstedt,
-Jeremy Jones, Chris Needham, Jer Noble, Justin Uberti, Yoav Weiss, and Eckhart
-Wörner for their contributions to this document.
+Jeremy Jones, Chris Needham, Jer Noble, Justin Uberti, Yoav Weiss, Eckhart
+Wörner and Benjamin Keen for their contributions to this document.
--- a/index.bs
+++ b/index.bs
@@ -481,12 +481,22 @@ regardless of the <a>browsing context</a>.
 
 ## Permissions Policy ## {#permissions-policy}
 
-This specification defines a <a>policy-controlled feature</a> named
-`"picture-in-picture"` that controls whether the <a>request Picture-in-Picture
-algorithm</a> may return a {{SecurityError}} and whether
-{{pictureInPictureEnabled}} is `true` or `false`.
+This specification defines two <a>policy-controlled features</a> that control
+whether picture-in-picture is allowed:
 
+* `"picture-in-picture"`.
+* `"browser-initiated-automatic-picture-in-picture"`.
+
+The <a>policy-controlled feature</a> named `"picture-in-picture"` controls
+whether the <a>request Picture-in-Picture algorithm</a> may return a
+{{SecurityError}} and whether {{pictureInPictureEnabled}} is `true` or `false`.
 The <a>default allowlist</a> for this feature is `*`.
+
+The <a>policy-controlled feature</a> named
+`"browser-initiated-automatic-picture-in-picture"` controls whether the
+<a>request Picture-in-Picture algorithm</a> may be invoked automatically by the
+<a>user agent</a>. When blocked by this policy, a console warning is recorded.
+The <a>default allowlist</a> for this feature is `self`.
 
 # Acknowledgments # {#acknowledgments}
 
--- a/index.bs
+++ b/index.bs
@@ -485,18 +485,17 @@ This specification defines two <a>policy-controlled features</a> that control
 whether picture-in-picture is allowed:
 
 * `"picture-in-picture"`.
-* `"browser-initiated-automatic-picture-in-picture"`.
+* `"automatic-picture-in-picture"`.
 
 The <a>policy-controlled feature</a> named `"picture-in-picture"` controls
 whether the <a>request Picture-in-Picture algorithm</a> may return a
 {{SecurityError}} and whether {{pictureInPictureEnabled}} is `true` or `false`.
 The <a>default allowlist</a> for this feature is `*`.
 
-The <a>policy-controlled feature</a> named
-`"browser-initiated-automatic-picture-in-picture"` controls whether the
-<a>request Picture-in-Picture algorithm</a> may be invoked automatically by the
-<a>user agent</a>. When blocked by this policy, a console warning is recorded.
-The <a>default allowlist</a> for this feature is `self`.
+The <a>policy-controlled feature</a> named `"automatic-picture-in-picture"`
+controls whether the <a>request Picture-in-Picture algorithm</a> may be invoked
+automatically by the <a>user agent</a>. When blocked by this policy, a console
+warning is recorded. The <a>default allowlist</a> for this feature is `self`.
 
 # Acknowledgments # {#acknowledgments}
 
--- a/index.bs
+++ b/index.bs
@@ -42,7 +42,6 @@ spec:dom; type:dfn; for:NamedNodeMap; text:element
 spec:dom; type:dfn; text:origin
 spec:dom; type:interface; text:Document
 spec:html; type:attribute; for:HTMLMediaElement; text:readyState
-spec:html; type:dfn; text:allowed to use
 spec:html; type:dfn; for:/; text:browsing context
 spec:infra; type:dfn; text:user agent
 spec:infra; type:dfn; for:list; text:item
@@ -193,30 +192,35 @@ the user agent MUST run the following steps:
 
 1. If <a>Picture-in-Picture support</a> is `false`, throw a
     {{NotSupportedError}} and abort these steps.
-2. If the document is not <a>allowed to use</a> the <a>policy-controlled feature</a>
+2. If the document is not [=/allowed to use=] the <a>policy-controlled feature</a>
     named `"picture-in-picture"`, throw a {{SecurityError}} and abort these
     steps.
-3. If |video|'s {{readyState}} attribute is {{HAVE_NOTHING}}, throw a
+3. If the document is not [=/allowed to use=] the <a>policy-controlled feature</a>
+    named `"automatic-picture-in-picture"` and the
+    {{MediaSessionEnterPictureInPictureReason}} enum is
+    {{MediaSessionEnterPictureInPictureReason/"contentoccluded"}}, throw a
+    {{SecurityError}} and abort these steps.
+4. If |video|'s {{readyState}} attribute is {{HAVE_NOTHING}}, throw a
     {{InvalidStateError}} and abort these steps.
-4. If |video| has no video track, throw a {{InvalidStateError}} and abort
+5. If |video| has no video track, throw a {{InvalidStateError}} and abort
     these steps.
-5. If |video|'s {{HTMLVideoElement/disablePictureInPicture}} is true,
+6. If |video|'s {{HTMLVideoElement/disablePictureInPicture}} is true,
     the user agent MAY throw an {{InvalidStateError}} and abort these steps.
-6. If {{pictureInPictureElement}} is `null` and the <a>relevant global object</a>
+7. If {{pictureInPictureElement}} is `null` and the <a>relevant global object</a>
     of <a>this</a> does not have <a>transient activation</a>, throw a
     {{NotAllowedError}} and abort these steps.
-7. If |video| is {{pictureInPictureElement}}, abort these steps.
-8. Set {{pictureInPictureElement}} to |video|.
-9. Let <dfn>Picture-in-Picture window</dfn> be a new instance of
+8. If |video| is {{pictureInPictureElement}}, abort these steps.
+9. Set {{pictureInPictureElement}} to |video|.
+10. Let <dfn>Picture-in-Picture window</dfn> be a new instance of
     {{PictureInPictureWindow}} associated with {{pictureInPictureElement}}.
-10. Append <a>relevant settings object</a>'s <a>origin</a> to
+11. Append <a>relevant settings object</a>'s <a>origin</a> to
     <a>initiators of active Picture-in-Picture sessions</a>.
-11. <a>Queue a task</a> to <a>fire an event</a> named
+12. <a>Queue a task</a> to <a>fire an event</a> named
     {{enterpictureinpicture}} using {{PictureInPictureEvent}} at the
     |video| with its {{bubbles}} attribute initialized to `true` and its
     {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
     <a>Picture-in-Picture window</a>.
-12. If {{pictureInPictureElement}} is <a>fullscreenElement</a>, it is
+13. If {{pictureInPictureElement}} is <a>fullscreenElement</a>, it is
     RECOMMENDED to <a>exit fullscreen</a>.
 
 It is RECOMMENDED that video frames are not rendered in the page and in the
@@ -374,7 +378,7 @@ partial interface Document {
 
 The {{pictureInPictureEnabled}} attribute's getter must return `true` if
 <a>Picture-in-Picture support</a> is `true` and <a>this</a> is
-<a>allowed to use</a> the feature indicated by attribute name
+[=/allowed to use=] the feature indicated by attribute name
 `picture-in-picture`, and `false` otherwise.
 
 <dfn>Picture-in-Picture support</dfn> is `false` if there's a user preference
@@ -493,26 +497,29 @@ whether picture-in-picture is allowed:
 <dl dfn-type=value>
   <dt>"<dfn permission export>picture-in-picture</dfn>".
   <dd>
-    Controls whether the [=document=] [=allowed to use|is allowed=] to use
+    Controls whether the [=document=] [=/allowed to use|is allowed=] to use
     Picture-in-Picture. The <a>default allowlist</a> for this feature is `*`.
 
   <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
   <dd>
-    Controls whether the user agent SHOULD automatically trigger
-    Picture-in-Picture. The <a>default allowlist</a> for this feature is
-    `self`.
+    Controls whether the [=document=] [=/allowed to use|is allowed=] to use
+    Picture-in-Picture when the {{MediaSessionEnterPictureInPictureReason}}
+    enum is {{MediaSessionEnterPictureInPictureReason/"contentoccluded"}}. The
+    <a>default allowlist</a> for this feature is `self`.
 </dl>
 
 Access to the Picture-in-Picture API is gated behind the
 <a>policy-controlled feature</a> {{picture-in-picture}}, which has a
 <a>default allowlist</a> of `*`.
 
-When the document is not [=allowed to use=] the
+When the document is not [=/allowed to use=] the
 <a>policy-controlled feature</a> named {{automatic-picture-in-picture}}, the
 user agent SHOULD not trigger a [[MediaSession]]
 {{MediaSessionAction/"enterpictureinpicture"}} action with the
-{{MediaSessionEnterPictureInPictureReason/"contentoccluded"}} reason. The
-<a>default allowlist</a> for this feature is `self`.
+{{MediaSessionEnterPictureInPictureReason/"contentoccluded"}} reason. If such
+action is triggered, the <a>request Picture-in-Picture algorithm</a>. will
+throw a {{SecurityError}}. The <a>default allowlist</a> for this feature is
+`self`.
 
 Note: User agents should surface a console warning whenever
 `"picture-in-picture"` is blocked by the {{automatic-picture-in-picture}}
--- a/index.bs
+++ b/index.bs
@@ -527,6 +527,6 @@ permissions policy.
 
 # Acknowledgments # {#acknowledgments}
 
-Thanks to Jennifer Apacible, Zouhir Chahoud, Marcos Cáceres, Philip Jägenstedt,
-Jeremy Jones, Chris Needham, Jer Noble, Justin Uberti, Yoav Weiss, Eckhart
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -481,13 +481,22 @@ regardless of the <a>browsing context</a>.
...
+* `"automatic-picture-in-picture"`.
+
+The <a>policy-controlled feature</a> named `"picture-in-picture"` controls
+whether the <a>request Picture-in-Picture algorithm</a> may return a
+{{SecurityError}} and whether {{pictureInPictureEnabled}} is `true` or `false`.
 The <a>default allowlist</a> for this feature is `*`.
 
+The <a>policy-controlled feature</a> named `"automatic-picture-in-picture"`
+controls whether the <a>request Picture-in-Picture algorithm</a> may be invoked
+automatically by the <a>user agent</a>. When blocked by this policy, a console
```
marcoscaceres, line 497: The console stuff seems also out of place as a statement of fact.

index.bs
```diff
@@ -481,13 +481,22 @@ regardless of the <a>browsing context</a>.
...
-{{pictureInPictureEnabled}} is `true` or `false`.
+This specification defines two <a>policy-controlled features</a> that control
+whether picture-in-picture is allowed:
 
+* `"picture-in-picture"`.
+* `"automatic-picture-in-picture"`.
+
+The <a>policy-controlled feature</a> named `"picture-in-picture"` controls
+whether the <a>request Picture-in-Picture algorithm</a> may return a
+{{SecurityError}} and whether {{pictureInPictureEnabled}} is `true` or `false`.
```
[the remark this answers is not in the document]
marcoscaceres: ah! lol, I see now... it was wrong in the first place... this is not well defined. 

We might need to fix this as a follow up.

marcoscaceres (changes requested):
The new policy needs to also be checked in the "Request Picture-in-Picture" algorithm.

index.bs
marcoscaceres, line 492: This is redundant... let's get rid of it, even though it was in the original spec. 

```suggestion
```

bpkeen:
> @bpkeen, who are you affiliated with? (initial IPR check failed)

Hi @marcoscaceres, I'm affiliated with @GoogleChrome, just updated my profile.

index.bs
bpkeen: This policy is still used and controls whether pip is allowed or not in general (due to user explicit request to enter pip or automatic behavior).

The additional policy I am proposing is to control the behavior for automatic requests to enter pip.

joeyparrish:
> Hi @marcoscaceres, I'm affiliated with @GoogleChrome, just updated my profile.

If you are at Google, could you please check out go/github and register so that internally, we can associate you with your GitHub account?  Thanks!

bpkeen:
> > Hi @marcoscaceres, I'm affiliated with @GoogleChrome, just updated my profile.
> 
> If you are at Google, could you please check out go/github and register so that internally, we can associate you with your GitHub account? Thanks!

Thanks @joeyparrish. Done!

index.bs
```diff
@@ -193,8 +197,8 @@ the user agent MUST run the following steps:
 1. If <a>Picture-in-Picture support</a> is `false`, throw a
     {{NotSupportedError}} and abort these steps.
 2. If the document is not <a>allowed to use</a> the <a>policy-controlled feature</a>
-    named `"picture-in-picture"`, throw a {{SecurityError}} and abort these
-    steps.
+    named {{picture-in-picture}} or {{automatic-picture-in-picture}}, throw a
```
beaufortfrancois, line 200: It's not clear to me why this algorithm should fail if website calls `video.requestPictureInPicture()` when the `automatic-picture-in-picture` feature is not allowed.

index.bs
```diff
@@ -303,6 +307,9 @@ Picture-in-Picture.
 The API will have to be used with the [[MediaSession]] API for customizing the
 available controls on the Picture-in-Picture window.
 
+The user agent SHOULD not override the <a>media session action</a>
+<a>enterpictureinpicture</a> action handler,  if one is provided by the site.
```
beaufortfrancois, line 311: Can you explain why to make it clear what is the intent of this addition?

beaufortfrancois (changes requested):

index.bs
bpkeen: You are correct, I should have removed this since the new policy does not apply when the website calls `video.requestPictureInPicture()`.

index.bs
```diff
@@ -42,6 +46,7 @@ spec:dom; type:dfn; for:NamedNodeMap; text:element
 spec:dom; type:dfn; text:origin
 spec:dom; type:interface; text:Document
 spec:html; type:attribute; for:HTMLMediaElement; text:readyState
+spec:html; type:dfn; text:allowed to use
```
marcoscaceres, line 49: Shouldn't need this. 

```suggestion
```

index.bs
```diff
@@ -303,6 +308,11 @@ Picture-in-Picture.
 The API will have to be used with the [[MediaSession]] API for customizing the
 available controls on the Picture-in-Picture window.
 
+The user agent SHOULD not override the <a>media session action</a>
+<a>enterpictureinpicture</a> action handler,  if one is provided by the site.
```
marcoscaceres, line 312: Instead of `<a>`, I think bikeshed uses `<{HTMLVideoElement/enterpictureinpicture}>` or just straight up `{{HTMLVideoElement/enterpictureinpicture}}`.

index.bs
```diff
@@ -481,15 +491,35 @@ regardless of the <a>browsing context</a>.
...
-`"picture-in-picture"` that controls whether the <a>request Picture-in-Picture
-algorithm</a> may return a {{SecurityError}} and whether
-{{pictureInPictureEnabled}} is `true` or `false`.
+This specification defines two <a>policy-controlled features</a> that control
+whether picture-in-picture is allowed:
+
+<dl dfn-type=value>
+  <dt>"<dfn permission export>picture-in-picture</dfn>".
+  <dd>
+    The `"picture-in-picture"` permission policy.
```
marcoscaceres, line 500: This is tautological, as it's not actually defining what the policy *does* and its default allow list. The definition here should state what the policy is for in a descriptive manner.  Or, alternatively (or concretely) link to the algorithm where this algorithm comes into play.

index.bs
```diff
@@ -481,15 +491,35 @@ regardless of the <a>browsing context</a>.
...
+<a>default allowlist</a> of `*`.
+
+When the document is not <a>allowed to use</a> the
+<a>policy-controlled feature</a> named {{automatic-picture-in-picture}}, the
+user agent SHOULD not trigger a [[MediaSession]] <a>enterpictureinpicture</a>
+action with the <a>contentoccluded</a> reason. The <a>default allowlist</a> for
+this feature is `self`.
 
-The <a>default allowlist</a> for this feature is `*`.
+Note: User agents should surface a console warning whenever
```
marcoscaceres, line 517: this note should be in the algorithm, where the allowed to use check is performed.

index.bs
```diff
@@ -481,15 +491,35 @@ regardless of the <a>browsing context</a>.
...
+  <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
+  <dd>
+    The `"automatic-picture-in-picture"` permission policy.
+</dl>
+
+Access to the Picture-in-Picture API is gated behind the
+<a>policy-controlled feature</a> {{picture-in-picture}}, which has a
+<a>default allowlist</a> of `*`.
+
+When the document is not <a>allowed to use</a> the
```
marcoscaceres, line 511: the "allowed to use" check needs to happen in an algorithm as part of an API call. It doesn't make sense here.

marcoscaceres (changes requested):
Thanks for making progress on this, @bpkeen! 

My comments are a bit rushed, so let me know if anything is unclear.

index.bs
```diff
@@ -481,15 +491,35 @@ regardless of the <a>browsing context</a>.
...
+    The `"picture-in-picture"` permission policy.
+
+  <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
+  <dd>
+    The `"automatic-picture-in-picture"` permission policy.
+</dl>
+
+Access to the Picture-in-Picture API is gated behind the
+<a>policy-controlled feature</a> {{picture-in-picture}}, which has a
+<a>default allowlist</a> of `*`.
```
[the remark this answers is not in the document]
marcoscaceres: I'm also confused... in the articles that @beaufortfrancois pointed to, if I understand correctly, it's only the top-level navigable that can initiate an automatic PiP? 

If so, what's the use case for the policy if not to enable it for third-party iframes?

index.bs
```diff
@@ -34,6 +34,10 @@ spec: Remote-Playback; urlPrefix: https://w3c.github.io/remote-playback/#dfn-
     type: dfn
         text: local playback device
         text: local playback state
+spec: MediaSession; urlPrefix: https://w3c.github.io/mediasession
```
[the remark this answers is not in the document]
bpkeen: Removed. Thank you for sharing the link (https://respec.org/xref/) very useful!

index.bs
bpkeen: If removed, `bikeshed spec` prints the following log:

Multiple possible 'allowed to use' dfn refs.
Arbitrarily chose https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:html; type:dfn; text:allowed to use
spec:private-aggregation-api; type:dfn; text:allowed to use
[=allowed to use=]

index.bs
bpkeen: AutoPiP is not part of the algorithm. The UA decides when to trigger the `enterpictureinpicture` action.

index.bs
bpkeen: Same as here https://github.com/w3c/picture-in-picture/pull/241#discussion_r2392868127. AutoPiP is an internal implementation detail of the UA.

index.bs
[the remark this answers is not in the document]
bpkeen: Yes, the automatic-picture-in-picture policy is defined with the `self` default allow list.

> if I understand correctly, it's only the top-level navigable that can initiate an automatic PiP?

Thats correct.

> If so, what's the use case for the policy if not to enable it for third-party iframes?

The purpose of the policy is to allow devs/site admins to control the automatic picture-in-picture behavior, without having to write code.

index.bs
```diff
@@ -481,15 +487,41 @@ regardless of the <a>browsing context</a>.
...
+    Controls whether the document is allowed to use Picture-in-Picture.
+    Specifically, the <a>request Picture-in-Picture algorithm</a> throws a
+    {{SecurityError}} if the document is not [=allowed to use=] this feature.
+    The <a>default allowlist</a> for this feature is `*`.
+
+  <dt>"<dfn permission export>automatic-picture-in-picture</dfn>".
+  <dd>
+    Controls whether the user agent SHOULD automatically trigger
+    Picture-in-Picture. The <a>default allowlist</a> for this feature is
+    `self`.
```
marcoscaceres, line 505: ok, but note "the <a>request Picture-in-Picture algorithm</a> throws" for "picture-in-picture". I was expecting the same thing here. When is this policy checked?

index.bs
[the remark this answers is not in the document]
marcoscaceres: I don't follow... only top-level navigable can set this, and it has no effect on iframes, then this policy serves no purpose. 

> The purpose of the policy is to allow devs/site admins to control the automatic picture-in-picture behavior, without having to write code.

But they have to write server code to set the HTTP header, right? So this doesn't hold. 

This breaks the invariant that permission policies apply to nested navigables. 

So, this is confusing:

```
<iframe
  src="https://third-party.com"
  allow="automatic-picture-in-picture">
</iframe>
``` 

If the above doesn't hold, then this is not a permissions policy.

marcoscaceres (changes requested):
### Summary of Points on `automatic-picture-in-picture` Permissions Policy

1. **Practical Usefulness Is Limited**
   - The policy defaults to `'self'`, so only the top-level document is affected.
   - It does not impact embedded content (iframes); for most sites, it "just works" by default.

2. **Developer Control Over API**
   - Developers are already in control: simply don't call the Picture-in-Picture API if PiP is not desired.
   - Explicitly setting the permissions policy to `"deny"` is redundant if the site doesn't invoke PiP programmatically.

3. **No Real Browser Auto-PiP**
   - User agents do not automatically trigger PiP without a user gesture or script.
   - The concept of UA-driven (automatic) PiP is speculative, right? 

4. **Ineffectiveness for Embeds/Iframes**
   - The policy does not apply to iframe content. Embedded media cannot be auto-PiP’d regardless of the policy setting. This violates the Permissions Policy contract/purpose. 

5. **Accessibility (a11y) Concerns**
   - If browsers ever auto-trigger PiP for accessibility reasons, developers should NOT be allowed to block this, as it would undermine user needs.
   - User accessibility requirements must take precedence over developer preferences.

6. **Edge Cases and Future-Proofing**
   - The only real value of the policy might be in future scenarios or as a defense-in-depth measure, but today it has little practical effect. Best defense is: don't write the code if you want to deny it in the first place.

index.bs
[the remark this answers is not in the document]
bpkeen: I was under the impression that admins could use tools to set these policies w/o the need of writing code.

To clarify, the above example will work, it is just that at the moment only the main frame is allowed to request PiP automatically. The result is that the given example would be a no-op.

index.bs
bpkeen: It would be checked by the browser, before attempting to request the PiP window.

bpkeen:
Thank you for the review Marcos. I've inlined responses to some of the points below.

I understand that there are concerns with the new policy as it stands. Are there any suggestions/modifications that you think could make this worth implementing? The biggest concern I have with not having the policy is that devs/admins will only be able to disable automatic PiP either through code or by blocking PiP in general. Which ultimately may be ok.

Looking forward to hearing your thoughts.

> ### Summary of Points on `automatic-picture-in-picture` Permissions Policy
> 1. **Practical Usefulness Is Limited**
>    
>    * The policy defaults to `'self'`, so only the top-level document is affected.
>    * It does not impact embedded content (iframes); for most sites, it "just works" by default.
> 2. **Developer Control Over API**
>    
>    * Developers are already in control: simply don't call the Picture-in-Picture API if PiP is not desired.

This is correct, this falls back to allowing an easy, policy driven way to prevent automatic PiP.

>    * Explicitly setting the permissions policy to `"deny"` is redundant if the site doesn't invoke PiP programmatically.
> 3. **No Real Browser Auto-PiP**
>    
>    * User agents do not automatically trigger PiP without a user gesture or script.
>    * The concept of UA-driven (automatic) PiP is speculative, right?

Could you please elaborate? There is ongoing development for this feature, it can be enabled with the following flag: `chrome://flags/#browser-initiated-automatic-picture-in-picture`

> 4. **Ineffectiveness for Embeds/Iframes**
>    
>    * The policy does not apply to iframe content. Embedded media cannot be auto-PiP’d regardless of the policy setting. This violates the Permissions Policy contract/purpose.
> 5. **Accessibility (a11y) Concerns**
>    
>    * If browsers ever auto-trigger PiP for accessibility reasons, developers should NOT be allowed to block this, as it would undermine user needs.
>    * User accessibility requirements must take precedence over developer preferences.

Thanks for bringing this up. However, even w/o this new policy sites can still block PiP using the `picture-in-picture` policy.

> 6. **Edge Cases and Future-Proofing**
>    
>    * The only real value of the policy might be in future scenarios or as a defense-in-depth measure, but today it has little practical effect. Best defense is: don't write the code if you want to deny it in the first place.

index.bs
marcoscaceres: Ok, I think you should now be able to just use [=/allowed to use=].

index.bs
bpkeen: I see where the gap is now. What I am proposing is a recommendation for the UA to not trigger the `enterpictureinpicture` to begin with. So the algorithm would not be called at all.

But your comment makes total sense, the policy should be tied to a concrete portion of the spec.

I've added the policy check to the request PiP algorithm and also kept the recommendation for UA's. Let me know what you think!

index.bs
```diff
@@ -192,30 +192,35 @@ the user agent MUST run the following steps:
 
 1. If <a>Picture-in-Picture support</a> is `false`, throw a
     {{NotSupportedError}} and abort these steps.
-2. If the document is not <a>allowed to use</a> the <a>policy-controlled feature</a>
+2. If the document is not [=/allowed to use=] the <a>policy-controlled feature</a>
     named `"picture-in-picture"`, throw a {{SecurityError}} and abort these
     steps.
-3. If |video|'s {{readyState}} attribute is {{HAVE_NOTHING}}, throw a
+3. If the document is not [=/allowed to use=] the <a>policy-controlled feature</a>
+    named `"automatic-picture-in-picture"` and the
```
marcoscaceres: Where did the enum come from? Was it from the document? 

I see there is some connection to media session, but it's not how this is checked

bpkeen: Yes, this comes from media session (see [here](https://www.w3.org/TR/mediasession/#enumdef-mediasessionenterpictureinpicturereason)). I've updated the section to clarify this.

index.bs
```diff
@@ -303,6 +308,11 @@ Picture-in-Picture.
 The API will have to be used with the [[MediaSession]] API for customizing the
 available controls on the Picture-in-Picture window.
 
+The user agent SHOULD not override the media session action
```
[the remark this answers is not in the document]
bpkeen: Done.

index.bs
```diff
@@ -303,6 +308,11 @@ Picture-in-Picture.
...
+{{MediaSessionAction/"enterpictureinpicture"}} action handler, if one is
+provided by the site. This gives the site control over the Picture-in-Picture
+user experience, when a Picture-in-Picture window is requested.
+
```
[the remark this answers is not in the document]
bpkeen: At the moment, I am drawing a blank. I can't think of any situations where overriding is warranted.

index.bs
```diff
@@ -481,15 +491,42 @@ regardless of the <a>browsing context</a>.
...
+  <dd>
+    Controls whether the [=document=] [=/allowed to use|is allowed=] to use
+    Picture-in-Picture when the {{MediaSessionEnterPictureInPictureReason}}
+    enum is {{MediaSessionEnterPictureInPictureReason/"contentoccluded"}}. The
+    <a>default allowlist</a> for this feature is `self`.
+</dl>
+
+Access to the Picture-in-Picture API is gated behind the
+<a>policy-controlled feature</a> {{picture-in-picture}}, which has a
+<a>default allowlist</a> of `*`.
```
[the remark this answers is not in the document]
bpkeen: Done, removed.

index.bs
```diff
@@ -192,30 +192,35 @@ the user agent MUST run the following steps:
...
 1. If <a>Picture-in-Picture support</a> is `false`, throw a
     {{NotSupportedError}} and abort these steps.
-2. If the document is not <a>allowed to use</a> the <a>policy-controlled feature</a>
+2. If the document is not [=/allowed to use=] the <a>policy-controlled feature</a>
     named `"picture-in-picture"`, throw a {{SecurityError}} and abort these
     steps.
-3. If |video|'s {{readyState}} attribute is {{HAVE_NOTHING}}, throw a
+3. If the document is not [=/allowed to use=] the <a>policy-controlled feature</a>
+    named `"automatic-picture-in-picture"` and the
+    {{MediaSessionEnterPictureInPictureReason}} enum is
```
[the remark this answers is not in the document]
bpkeen: Please see [this comment](https://github.com/w3c/picture-in-picture/pull/241/#discussion_r2876054500).

index.bs
```diff
@@ -481,15 +491,42 @@ regardless of the <a>browsing context</a>.
...
+    <a>default allowlist</a> for this feature is `self`.
+</dl>
+
+Access to the Picture-in-Picture API is gated behind the
+<a>policy-controlled feature</a> {{picture-in-picture}}, which has a
+<a>default allowlist</a> of `*`.
+
+When the document is not [=/allowed to use=] the
+<a>policy-controlled feature</a> named {{automatic-picture-in-picture}}, the
+user agent SHOULD not trigger a [[MediaSession]]
```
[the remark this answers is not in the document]
bpkeen: Updated.

index.bs
```diff
@@ -192,30 +192,35 @@ the user agent MUST run the following steps:
 
 1. If <a>Picture-in-Picture support</a> is `false`, throw a
     {{NotSupportedError}} and abort these steps.
-2. If the document is not <a>allowed to use</a> the <a>policy-controlled feature</a>
+2. If the document is not [=/allowed to use=] the <a>policy-controlled feature</a>
```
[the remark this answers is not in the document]
bpkeen: This is much better, thanks!

index.bs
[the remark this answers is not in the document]
marcoscaceres: Should we remove it or can you think of anyone else we can ask? @beaufortfrancois maybe?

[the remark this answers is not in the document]
beaufortfrancois: The only theoretical way I could think of is if a site’s custom handler intentionally prevents the user from entering Picture-in-Picture like forcing the user to stay on the page to see ads for instance.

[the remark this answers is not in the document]
bpkeen: Since we don't have real world examples, I think we could keep the note generalized and focus on what is best for the user. How about this:

> Note: The user agent is generally expected to prioritize a site’s registered media session action handler. However, the user agent retains the discretion to override this behavior if it determines that such an intervention would result in an overall improved user experience. This is particularly relevant:
> 
> - In cases of perceived abuse, such as a site intentionally blocking Picture-in-Picture to force user attention on advertisements.
> - When fulfilling explicit user preferences and/or accessibility requirements that the site's implementation may fail to meet.

index.bs
```diff
@@ -30,6 +30,9 @@ spec: Fullscreen; urlPrefix: https://fullscreen.spec.whatwg.org
     type: dfn; text: fullscreen flag; url: #fullscreen-flag
     type: dfn; text: fullscreenElement; url: dom-document-fullscreenelement
     type: dfn; text: exit fullscreen; url: fully-exit-fullscreen
+spec: MediaSession; urlPrefix: https://w3c.github.io/mediasession/
```
marcoscaceres: Are you sure you need these? These should be unique and BikeShed should find them?

marcoscaceres:
Just a heads up that this will have a lot of merge conflicts with https://github.com/w3c/picture-in-picture/pull/245

Do we have a second implementer willing to support this?

index.bs
```diff
@@ -188,31 +191,37 @@ the user agent MUST run the following steps:
...
-2. If the document is not <a>allowed to use</a> the <a>policy-controlled feature</a>
+1. Let |document| be [=this=]'s [=/relevant global object=]'s [=/associated Document=].
+1. If the |document| is not [=/allowed to use=] the [=policy-controlled feature=]
     named `"picture-in-picture"`, throw a {{SecurityError}} and abort these
     steps.
-3. If |video|'s {{readyState}} attribute is {{HAVE_NOTHING}}, throw a
+1. If the |document| is not [=/allowed to use=] the <a>policy-controlled feature</a>
+    named `"automatic-picture-in-picture"` and the
+    {{MediaSessionEnterPictureInPictureReason}} enum, defined in [[MediaSession]], is
+    {{MediaSessionEnterPictureInPictureReason/"contentoccluded"}}, throw a
```
jan-ivar: This makes no sense. An `enum` is not a value, and cannot be read.

I see no value passed into this algorithm this could be referring to.

The only argument passed into this algorithm is _video_, from which we find a global and a document, but not a media session let alone any specific media session action holding [details](https://w3c.github.io/mediasession/#dom-mediasessionactionhandler-details) AFAICT.

If the intent here is to detect JS calling this method like this:
```js
navigator.mediaSession.setActionHandler("enterpictureinpicture", () =>
    video.requestPictureInPicture());
```
...as distinct from e.g.
```js
button.onclick = () => video.requestPictureInPicture();
```
...then how this is to be done must be specified and cannot be inferred from closure.

How is this implemented?

Should the following pass or fail?
```js
navigator.mediaSession.setActionHandler("enterpictureinpicture", async () => {
  await undefined;
  video.requestPictureInPicture();
});
```

jan-ivar (changes requested):

index.bs
```diff
@@ -467,15 +483,39 @@ regardless of the <a>browsing context</a>.
...
+</dl>
+
+When the document is not [=/allowed to use=] the
+<a>policy-controlled feature</a> named {{automatic-picture-in-picture}}, the
+user agent SHOULD NOT trigger a [[MediaSession]]
+{{MediaSessionAction/"enterpictureinpicture"}} action with the
+{{MediaSessionEnterPictureInPictureReason/"contentoccluded"}} reason. If such
+action is triggered, the <a>request Picture-in-Picture algorithm</a>. will
+throw a {{SecurityError}}. The <a>default allowlist</a> for this feature is
+`self`.
```
jan-ivar: First, thank you for adding this, as it's important every new [powerful feature](https://www.w3.org/TR/permissions/#dfn-powerful-feature) is specified somewhere as to how it's supposed to work.

But this is functionality gated by _permission_ (prompts etc.), which is distinct from _permission policy_ (iframe delegation). It needs its own section.

To clarify, specs that define powerful features (e.g [MediaCapture-main](https://w3c.github.io/mediacapture-main)) need _two_ sections:
1. [§ 13. Permissions Integration](https://w3c.github.io/mediacapture-main/#permissions-integration) defining the [_powerful feature_](https://www.w3.org/TR/permissions/#dfn-powerful-feature), e.g. `"camera"`
2. [§ 14. Permissions Policy Integration](https://w3c.github.io/mediacapture-main/#permissions-policy-integration) defining the [_policy-controlled feature_](https://www.w3.org/TR/permissions-policy-1/#policy-controlled-feature), e.g. also `"camera"`

The former is for use with permission algorithms like [request permission to use](https://www.w3.org/TR/permissions/#dfn-request-permission-to-use).
The latter is for use with iframe policy algorithms like [allowed to use](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use) (yes it's confusing).

In your case you already define `picture-in-picture` and `automatic-picture-in-picture` correctly for the latter.

But you need to also define `automatic-picture-in-picture` as a permission (prompt) gated powerful feature, and how it works (in contrast `picture-in-picture` has no permission or prompt).

The section you have here comes close to describing how the powerful feature works. It's unusual in that the powerful feature isn't directly anchored to any API method (except perhaps [setActionHandler](https://w3c.github.io/mediasession/#dom-mediasession-setactionhandler)), but I think we still need to define how (or when in this case) it should be expected to work.

As such, the above description appears to differ from how the Chrome blog [Automatic picture-in-picture for video conferencing web apps](https://developer.chrome.com/blog/automatic-picture-in-picture#enter_picture-in-picture_automatically) describes it:
> A web app is only eligible for automatic picture-in-picture if it meets all of the following conditions:
> 
> 1. It has registered a [media session action](https://www.w3.org/TR/mediasession/#ref-for-enumdef-mediasessionaction) handler for the "enterpictureinpicture" action.
> 
> 2. It is actively capturing camera or microphone using [getUserMedia](https://developer.mozilla.org/docs/Web/API/MediaDevices/getUserMedia).
> 
> 3. The user allows "automatic picture-in-picture" through a browser setting enabled by default.

Mozilla is interested in implementing this feature, but wants to make sure it will be web compatible.
cc @youennf

jan-ivar (changes requested):

index.bs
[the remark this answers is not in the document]
chrisn: Note that per https://github.com/w3c/picture-in-picture/issues/191 we should move this section into the Media Session spec. That spec describes default handlers for other actions, so we can do similar there for `enterpictureinpicture`. For other Media Session actions, the spec says that the UA may (or is recommended) to provide a handler if none is provided by the site. It's not clear to me why `enterpictureinpicture` is different.

index.bs
bpkeen: I've removed them.

index.bs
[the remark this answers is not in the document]
bpkeen: Ackd, I have removed this section.

index.bs
bpkeen: You are right, to resolve this I've refactored the spec to formalize "Automatic Picture-in-Picture request". Please note that it is UA driven. Now the `requestPictureInPicture()` do not reference the enum.

index.bs
bpkeen: Thank you for the feedback Jan-Ivar. I have added a `Permissions Integration` section.

Regarding the blog, I've added a note to the spec clarifying that the UA may enforce additional eligibility requirements. Please let me know if you have any additional feedback.

index.bs
```diff
@@ -476,17 +483,71 @@ The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
...
 regardless of the <a>browsing context</a>.
 
+## Permissions Integration ## {#permissions-integration}
+
+This specification defines a <a>powerful feature</a> identified by the
+[=powerful feature/name=] `"automatic-picture-in-picture"`.
+
+The `"automatic-picture-in-picture"` permission controls access to the
+`"automatic-picture-in-picture"` powerful feature.
+
```
jan-ivar: My understanding is the powerful feature [name](https://w3c.github.io/permissions/#dfn-name) already _is_ the [permission name](https://w3c.github.io/permissions/#ref-for-dfn-name-2), so this seems circular.
```suggestion
```

index.bs
```diff
@@ -476,17 +483,71 @@ The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
...
+The user agent MUST NOT initiate an [=automatic Picture-in-Picture request=]
+for a {{Document}} if the {{Document}}'s [=relevant settings object=]'s
+[=permission state=] for the `"automatic-picture-in-picture"` feature is
+{{PermissionState/"denied"}}.
+
+If the [=permission state=] is {{PermissionState/"prompt"}}:
+1. The user agent MUST prompt the user for permission upon initiating the [=automatic Picture-in-Picture request=].
```
jan-ivar: The permission spec has algorithms we can use here. 
```suggestion
1. The user agent MUST [=request permission to use=] `"automatic-picture-in-picture"` before initiating an [=automatic Picture-in-Picture request=].
```

index.bs
```diff
@@ -476,17 +483,71 @@ The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
...
+2. If the user denies the permission, the user agent MUST run the [=automatic Picture-in-Picture permission revocation algorithm=].
+
+Note: When the [=permission state=] is {{PermissionState/"prompt"}}, user
+agents should display the prompt within the Picture-in-Picture window, blocking
```
jan-ivar: Follow https://w3c.github.io/picture-in-picture/#w3c-conventions

```suggestion
agents SHOULD display the prompt within the Picture-in-Picture window, blocking
```

index.bs
```diff
@@ -476,17 +483,71 @@ The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
...
+all other interaction until the user responds.
+
+If the [=permission state=] is {{PermissionState/"granted"}}, the user agent
+may initiate the [=automatic Picture-in-Picture request=] without prompting.
```
jan-ivar: We probably want something stronger than MAY here right?
```suggestion
SHOULD initiate the [=automatic Picture-in-Picture request=] without prompting.
```

index.bs
```diff
@@ -476,17 +483,71 @@ The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
...
+
+Note: The user agent may enforce additional eligibility criteria before
+initiating an [=automatic Picture-in-Picture request=].
```
jan-ivar: Notes are informative, so normative statements in notes are confusing.

```suggestion
The user agent MAY enforce additional eligibility criteria before
initiating an [=automatic Picture-in-Picture request=].
```

index.bs
```diff
@@ -476,17 +483,71 @@ The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
...
+    feature is `self`.
+</dl>
+
+If the {{Document}} is not [=allowed to use=] <a permission>automatic-picture-in-picture</a>,
+the user agent MUST NOT initiate [=automatic Picture-in-Picture requests=].
 
-The <a>default allowlist</a> for this feature is `*`.
+Note: User agents should surface a console warning whenever
+an [=automatic Picture-in-Picture request=] is blocked by the
+<a permission>automatic-picture-in-picture</a> permissions policy.
```
jan-ivar:
```suggestion
Note: User agents are encouraged to surface a web console warning whenever
an [=automatic Picture-in-Picture request=] is blocked by the
<a permission>automatic-picture-in-picture</a> permissions policy.
```

jan-ivar (changes requested):

index.bs
bpkeen: Done

index.bs
bpkeen: Thanks for the pointer. I've decided to remove this note, as to not constrain user agent's implementations.

index.bs
bpkeen: I've reworded this to:

```
When initiating an [=automatic Picture-in-Picture request=], the user agent
MUST [=request permission to use=] `"automatic-picture-in-picture"`.
```

My concern with running the `request permission to use` algorithm **before** initiating the request is that it can be confusing to users, since this would take place during occlusion. Please let me know what you think.

index.bs
```diff
@@ -193,6 +193,15 @@ An origin is said to have an active Picture-in-Picture session if any
 of the origins in <a>initiators of active Picture-in-Picture sessions</a>
 are <a>same origin-domain</a> with origin.
 
+An <dfn export>automatic Picture-in-Picture request</dfn> is a request to enter
+Picture-in-Picture initiated by the user agent due to content occlusion. This
```
marcoscaceres, line 197: This sentence is confusing... what does " due to content occlusion" mean?

index.bs
```diff
@@ -476,17 +483,58 @@ The API is not limited to [[SECURE-CONTEXTS]] because it exposes a feature
...
+    Controls whether the [=powerful feature=] `"automatic-picture-in-picture"`
+    is permitted in the {{Document}}. The <a>default allowlist</a> for this
+    feature is `self`.
+</dl>
+
+If the {{Document}} is not [=allowed to use=] <a permission>automatic-picture-in-picture</a>,
+the user agent MUST NOT initiate [=automatic Picture-in-Picture requests=].
 
-The <a>default allowlist</a> for this feature is `*`.
+Note: User agents are encouraged to surface a web a console warning whenever
```
tidoust, line 532:
```suggestion
Note: User agents are encouraged to surface a console warning whenever
```

index.bs
```diff
@@ -321,9 +330,8 @@ The {{requestPictureInPicture()}} method steps <dfn export>request Picture-in-Pi
 4. If [=this=]'s {{readyState}} attribute is {{HAVE_NOTHING}}, return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
 5. If [=this=] has no video track, return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
 6. If [=this=]'s {{HTMLVideoElement/disablePictureInPicture}} is true, user agent may return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
-7. If |doc|'s [=Picture-in-Picture element=] is `null`:
-    1. If [=this=]'s [=relevant global object=] does not have a [=transient activation=], then return [=a
-       promise rejected with=] {{NotAllowedError}} {{DOMException}}.
+7. If |doc|'s [=Picture-in-Picture element=] is `null` and the user agent is not executing these steps directly as part of an [=automatic Picture-in-Picture request=]:
```
marcoscaceres, line 333: There need to be some kind of flag that marks a Pip as "automatic Picture-in-Picture request"

index.bs
marcoscaceres: Oh! We figured it out... I think... if the video gets scrolled off the page? I guess it's sorta, like if this video was playing in a page and then it detaches: 

Image


But that seems very bad (specially if it's an ad)... it's still the user's choice if it should go into PiP, not the web page. 

The proposal as written makes it sound like "the user agent" is doing it... but it's the page getting the user agent to act on it's behalf, which is not in the interest of the user.

index.bs
marcoscaceres: This is also skipping user activation.

I don't think this should ever be allowed, to be honest... the right thing is for the "PiP" to happen in the page by detaching the video element, and then allowing the user to choose if they **really** want to PiP a video.

marcoscaceres:
It might be worth while filing a standards position for this on both Mozilla and WebKit to get clarity if there's any interest in moving this proposal forward. 

Alternatively, we should discuss this on a working group call to get views from the working group.

index.bs
bpkeen: Our implementation triggers the PiP request when the user switches tabs, with ongoing work to expand it window occlusion (a window occluded by another window).

> The proposal as written makes it sound like "the user agent" is doing it

To clarify, the user agent controls when to trigger, while the page is only allowed to customize the experience. If the page has an action handler, then the user agent triggers that, since the site wants to control the PiP experience. When there is no action handler the user agent will select a video to enter PiP.

For both cases the user agent is triggering the action, which allows enforcing additional eligibility requirements, to ensure the window opens when it is helpful.

index.bs
bpkeen: > the right thing is for the "PiP" to happen in the page

The issue here is that PiP triggers on content occlusion, which includes switching to a new tab, or a window covering another. PiP happening "in the page" would provide no value for the user since they can't see it.

I believe you are thinking of content occlusion by scrolling the video off the viewport, please correct me if I'm wrong. But that is just a subset, and actually a case I had not thought of.

bpkeen:
> It might be worth while filing a standards position for this on both Mozilla and WebKit to get clarity if there's any interest in moving this proposal forward.
> 
> Alternatively, we should discuss this on a working group call to get views from the working group.

That makes sense. Thanks @jan-ivar for filing the Mozilla standards position (https://github.com/mozilla/standards-positions/issues/1434). I've just filed the WebKit one here: https://github.com/WebKit/standards-positions/issues/707

index.bs
jan-ivar: I think the compelling use case for us is [in video conferencing](https://developer.chrome.com/blog/automatic-picture-in-picture).

Navigating tabs while on a call is a form of multi-tasking on the part of the user. An automatic "Meeting PiP" can be helpful to users to remind them they're still on an active call, as well as provide them a simple way to maintain or get back to the call.

While it's generally hard for a browser to 100% detect that video conferencing is happening, user agents can make inferences from camera and microphone being active. Camera and microphone sharing also indicate a high level of trust from the user, so there seems to be potential here to restrict auto pip to these narrow situations.

PR #253, opened 2026-05-11
pictureInPictureElement algorithm fix

### LINKED ISSUE #248: Picture in Picture getter issues/needs clarification
The getter steps are as follows: https://w3c.github.io/picture-in-picture/#dom-documentorshadowroot-pictureinpictureelement:

The [pictureInPictureElement](https://w3c.github.io/picture-in-picture/#dom-documentorshadowroot-pictureinpictureelement) attribute’s getter must run these steps:
1. If [this](https://webidl.spec.whatwg.org/#this) is a [shadow root](https://dom.spec.whatwg.org/#concept-shadow-root) and its [host](https://dom.spec.whatwg.org/#concept-documentfragment-host) is not [connected](https://dom.spec.whatwg.org/#connected), return null and abort these steps.
2. Let candidate be the result of [retargeting](https://dom.spec.whatwg.org/#retarget) Picture-in-Picture element against [this](https://webidl.spec.whatwg.org/#this).
3. If candidate and [this](https://webidl.spec.whatwg.org/#this) are in the same [tree](https://dom.spec.whatwg.org/#concept-tree), return candidate and abort these steps.
4. Return null.

But [tests like these](https://github.com/web-platform-tests/wpt/blob/406144885e9034efab057d7931023532fb120719/picture-in-picture/request-picture-in-picture-twice.html#L22-L26) we can see that we create a video, never binds it.

This means that the video's tree is 1 element long (containing itself), and therefore the candidate is not "in the same tree as this" (the document).

However, Chrome passes this, so it's pretty likely chrome has a bug wrt retargeting (or maybe chrome just skip step 3 alltogether? I don't know), but that's a separate issue. We must address the tests or the spec, as they are in conflict with each other.

cc @beaufortfrancois

theIDinside:
Addresses the issue with the getter algorithm and the behavior that two major browsers have, diverging from spec, when considering unbound/disconnected elements, to be in the same tree as the owner document.

Web platform tests account for this workiing, and as the test page in the issue provides an example that also works in safari.

The counter argument would be; "Why would we want to PIP disconnected elements" and the reason I could think of was, like a dynamic site that removes elements as the user scrolls and what not, and the video should still be playable (and reachable via `document.pictureInPictureElement`).

But the spec algorithm itself, doesn't actually allow for this behavior, without a fix like this one.

@chrisn @marcoscaceres @beaufortfrancois 

Closes #248

***
<a href="https://pr-preview.s3.amazonaws.com/theIDinside/picture-in-picture/pull/253.html" title="Last updated on May 26, 2026, 9:34 AM UTC (7c6e14b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/253/9d88d7b...theIDinside:7c6e14b.html" title="Last updated on May 26, 2026, 9:34 AM UTC (7c6e14b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -391,12 +391,16 @@ partial interface mixin DocumentOrShadowRoot {
 
 The {{pictureInPictureElement}} attribute's getter must run these steps:
 
-1. If <a>this</a> is a <a for=/>shadow root</a> and its <a for=DocumentFragment>host</a>
-    is not <a>connected</a>, return `null` and abort these steps.
-2. Let |candidate| be the result of [=retargeting=] [=Picture-in-Picture element=] against [=this=].
-3. If |candidate| and <a>this</a> are in the same <a>tree</a>,
-    return |candidate| and abort these steps.
-4. Return `null`.
+1. If [=this=] is a <a for=/>shadow root</a> and its <a for=DocumentFragment>host</a>
+    is not <a>connected</a>, return `null`.
+2. Let |candidate| be the result of <a>retargeting</a> Picture-in-Picture
+    element against [=this=].
+3. If |candidate| is `null`, return `null`.
+4. If |candidate| and [=this=] are in the same [=tree=],
+    return |candidate|.
+5. If [=this=] is a {{Document}} and |candidate|'s [=node document=]
+    is [=this=], return |candidate|.
+6. Return `null`.
 
 ## Interface <code>PictureInPictureWindow</code> ## {#interface-picture-in-picture-window}
 
--- a/index.bs
+++ b/index.bs
@@ -400,14 +400,14 @@ The {{pictureInPictureElement}} attribute's getter must run these steps:
     return |candidate|.
 5. If [=this=] is a {{Document}} and |candidate|'s [=node document=]
     is [=this=], return |candidate|.
-6. Return `null`.
 
-Note: Step 5 handles the case where |candidate| is not <a>connected</a> to
-<a>this</a>'s <a>tree</a> but its <a>node document</a> is still <a>this</a>.
-For example, an element that was created via {{Document/createElement()}}
-but never inserted into the tree. In that case |candidate| and <a>this</a>
-are not in the same <a>tree</a> (so step 4 does not match), but the element
-still belongs to <a>this</a> Document and is returned to script.
+    Note: This handles the case where |candidate| is not [=connected=] to
+    [=this=]'s [=tree=] but its [=node document=] is still [=this=].
+    For example, an element that was created via {{Document/createElement()}}
+    but never inserted into the tree. In that case |candidate| and [=this=]
+    are not in the same [=tree=] (so step 4 does not match), but the element
+    still belongs to [=this=] Document and is returned to script.
+6. Return `null`.
 
 ## Interface <code>PictureInPictureWindow</code> ## {#interface-picture-in-picture-window}
 
--- a/index.bs
+++ b/index.bs
@@ -402,6 +402,13 @@ The {{pictureInPictureElement}} attribute's getter must run these steps:
     is [=this=], return |candidate|.
 6. Return `null`.
 
+Note: Step 5 handles the case where |candidate| is not <a>connected</a> to
+<a>this</a>'s <a>tree</a> but its <a>node document</a> is still <a>this</a>.
+For example, an element that was created via {{Document/createElement()}}
+but never inserted into the tree. In that case |candidate| and <a>this</a>
+are not in the same <a>tree</a> (so step 4 does not match), but the element
+still belongs to <a>this</a> Document and is returned to script.
+
 ## Interface <code>PictureInPictureWindow</code> ## {#interface-picture-in-picture-window}
 
 <pre class="idl">


index.bs
```diff
@@ -381,12 +381,22 @@ partial interface mixin DocumentOrShadowRoot {
...
 2. Let |candidate| be the result of <a>retargeting</a> Picture-in-Picture
     element against <a>this</a>.
-3. If |candidate| and <a>this</a> are in the same <a>tree</a>,
-    return |candidate| and abort these steps.
-4. Return `null`.
+3. If |candidate| is null, return null.
+4. If |candidate| and <a>this</a> are in the same <a>tree</a>,
+    return |candidate|.
+5. If <a>this</a> is a Document and |candidate|'s <a>node document</a>
+    is this, return candidate.
```
chrisn:
```suggestion
    is <a>this</a>, return |candidate|.
```

theIDinside:
Even with this fix, the algorithm still has surprising effects for Shadow DOM stuff. Because of step 1:

1. If [this](https://webidl.spec.whatwg.org/#this) is a [shadow root](https://dom.spec.whatwg.org/#concept-shadow-root) and its [host](https://dom.spec.whatwg.org/#concept-documentfragment-host) is not [connected](https://dom.spec.whatwg.org/#connected), return null.

And because of the change of this PR (to reflect behavior of being able to PIP unbound elements), step 1 means that the following assert should always be true:

```js
// create an unbound `video` element
const video = await loadVideo();
const host = document.createElement('div');
const shadow = host.attachShadow({ mode });

await requestPictureInPictureWithTrustedClick(video);

// Unbound elements, via retargeting returns `host`
assert_equals(document.pictureInPictureElement, host);

// Step 1 of the algorithm however should return null here
assert_equals(shadow.pictureInPictureElement, null);
```

But given that blink and webkit behaves the way they do, they would fail on that last assert.

Should we remove step 1 from the algorithm? I think that would solve this?

marcoscaceres:
@theIDinside would you mind rebasing?

index.bs
```diff
@@ -381,12 +381,22 @@ partial interface mixin DocumentOrShadowRoot {
...
 2. Let |candidate| be the result of <a>retargeting</a> Picture-in-Picture
     element against <a>this</a>.
-3. If |candidate| and <a>this</a> are in the same <a>tree</a>,
-    return |candidate| and abort these steps.
-4. Return `null`.
+3. If |candidate| is null, return null.
+4. If |candidate| and <a>this</a> are in the same <a>tree</a>,
+    return |candidate|.
+5. If <a>this</a> is a Document and |candidate|'s <a>node document</a>
+    is <a>this</a>, return |candidate|.
```
```suggestion
4. If |candidate| and [=this=] are in the same [=tree=],
    return |candidate|.
5. If [=this=] is a {{Document}} and |candidate|'s [=node document=]
    is [=this=], return |candidate|.
```

index.bs
```diff
@@ -381,12 +381,22 @@ partial interface mixin DocumentOrShadowRoot {
...
+6. Return `null`.
+
+Note: Step 5 handles the case where |candidate| is not <a>connected</a> to
```
marcoscaceres: Nit. It might be better to move this note up into step 5. Also just minor bikeshed liking things as above.

index.bs
marcoscaceres: For "are in the same tree", please refer to https://dom.spec.whatwg.org/#concept-tree ... it lists a bunch of useful definitions for how to precisely check "are in the same tree".

marcoscaceres (commented):
This looks ok, but didn't get a chance to verify against WebKit's code. 

Will do that after the rebase so I can compare against the latest spec text in context.

index.bs
```diff
@@ -391,12 +391,23 @@ partial interface mixin DocumentOrShadowRoot {
...
 The {{pictureInPictureElement}} attribute's getter must run these steps:
 
-1. If <a>this</a> is a <a for=/>shadow root</a> and its <a for=DocumentFragment>host</a>
-    is not <a>connected</a>, return `null` and abort these steps.
-2. Let |candidate| be the result of [=retargeting=] [=Picture-in-Picture element=] against [=this=].
-3. If |candidate| and <a>this</a> are in the same <a>tree</a>,
-    return |candidate| and abort these steps.
-4. Return `null`.
+1. If [=this=] is a <a for=/>shadow root</a> and its <a for=DocumentFragment>host</a>
+    is not <a>connected</a>, return `null`.
```
marcoscaceres, line 395: Picking up @theIDinside's shadow-root point from #248: with step 5, a disconnected element is now returned for a `Document`, but step 1 still returns `null` for a disconnected shadow root even when the Picture-in-Picture element lives inside that shadow tree. That asymmetry is real, but I think it's separate from this PR. Could we land this for the `Document` case and open a dedicated issue to decide whether step 1 should be removed or relaxed to match the `Document` behaviour?

index.bs
```diff
@@ -391,12 +391,23 @@ partial interface mixin DocumentOrShadowRoot {
...
+2. Let |candidate| be the result of <a>retargeting</a> Picture-in-Picture
+    element against [=this=].
```
marcoscaceres, line 397: Linking nit: this drops the autolinks that are on `main` (`[=retargeting=] [=Picture-in-Picture element=]`), so "Picture-in-Picture element" no longer links to its definition. Suggest restoring them:

```suggestion
2. Let |candidate| be the result of [=retargeting=] [=Picture-in-Picture element=] against [=this=].
```

marcoscaceres (commented):
Thanks for the rebase. Now that it's against current `main`, I went and verified the getter against WebKit's implementation; overall I'm happy with the direction. A few specifics inline.

One correction to the example in #248: retargeting an unbound, unrelated video against the document returns the video, not the host (the video's root is not a shadow root, so `retarget` returns it directly), so `document.pictureInPictureElement` there would be the video rather than `host`.

index.bs
```diff
@@ -391,12 +391,23 @@ partial interface mixin DocumentOrShadowRoot {
...
+3. If |candidate| is `null`, return `null`.
+4. If |candidate| and [=this=] are in the same [=tree=],
+    return |candidate|.
+5. If [=this=] is a {{Document}} and |candidate|'s [=node document=]
+    is [=this=], return |candidate|.
```
marcoscaceres, line 402: I verified this against WebKit. An element created with `createElement()` has its tree scope set to its `Document` before it is ever inserted, so the getter retargets and returns it (`TreeScope::ancestorElementInThisScope`). Combined with `requestPictureInPicture()` having no "connected" check (it sets the document's Picture-in-Picture element to `this` even when unbound), the getter does need to return it here. The existing WPT `picture-in-picture-element.html` relies on exactly this: `loadVideo()` creates the video and never inserts it, then asserts `document.pictureInPictureElement === video`. So step 5 is the internally consistent fix.

marcoscaceres:
Following up on the step 1 / shadow-root point with implementation data on both engines.

WebKit just landed a fix to its `pictureInPictureElement` getter (WebKit/WebKit#67167, `315403@main`, merged 2026-06-17): the shadow-host connectivity check was inverted. With the fix, WebKit returns `null` only when the shadow root's host is not connected — i.e. it now follows step 1. Its regression test covers the connected case (the getter returns the retargeted host/video), and the corrected check returns `null` for a disconnected host.

Chrome differs on the disconnected case: for a video in a disconnected shadow tree, `shadowRoot.pictureInPictureElement` returns the video rather than `null` (which is what `wpt picture-in-picture/shadow-dom.html` asserts it should be). Chrome does follow step 5 for the document getter — `document.pictureInPictureElement` returns the host.

So step 1 for shadow roots is now an interop question rather than a hypothetical: WebKit has just implemented it as written, while Chrome returns the element. Worth weighing before removing or relaxing step 1 — it leans toward keeping it, with the Document-vs-shadow-root asymmetry tracked separately.

Issue #252, opened 2026-03-10
Exit Picture In Picture Algorithm Overhaul


theIDinside:
Adding this issue to track the `document.exitPictureInPicture()` request, as it needs the same kind of overhaul that the `requestPictureInPicture()` algorithm does (see #245)

I *think* it may be solved by just wrapping the whole thing inside a queued task. It still needs to probably be in-parallel, given that we don't really know/can't specify how long "closing the PIP window" will take, or in how many disjoint/asynchronous steps that actually happen inside the browser implementations. It could be 1, it could be many.

Edit: I will be adding things that need to change as I come across them, thinking about this algorithm.

chrisn:
As part of this, we should update the [unloading document cleanup steps](https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-document-cleanup-steps) in HTML to call to the Exit Picture In Picture algorithm.

theIDinside:
When this algorithm gets updated, `requestPictureInPicture` algorithm may also needs updating 11.5.1 `If {{pictureInPictureElement}} is not `null`, run the [=exit Picture-in-Picture algorithm=].`. But that's if this algorithm changes to take input variables instead of just accessing stuff, see [comment](https://github.com/w3c/picture-in-picture/pull/245#issuecomment-4397725387). (This may or may not change though, just left here for posterity)

theIDinside:
Given that `requestPictureInPicture` will call exit algorithm steps (see #245 PR) the [Queue a task to fire ...](https://w3c.github.io/picture-in-picture/#exit-pip:~:text=pictureInPictureElement%2E-,Queue,pictureInPictureElement%2E), which is step 3, must stop doing so.

Queuing a task is an abstraction to allow for DOM mutations, but it also has effects on ordering and timing. If we're already running on a queued task "in line", queuing another task is unnecessary. It also means that, for the request case, the ordering of the events will be:

- video1 requests & enters pip
- video 2 requests pip
- request algorithm sees that video1 is pip, exit pip for video1, queue a task to fire event
- video 2 enters pip and fires event

Which means that video2 sees it's enter event before video1's leave event is observed. 

Suggestion: Let's be more exact with how and where we queue tasks. This algorithm should run in-line with no queuing of tasks and instead let that be the responsibility of the caller, to make sure it's wrapped appropriately.

theIDinside:
[...]

PR #245, opened 2026-02-06
Address requestPictureInPicture() algorithm

theIDinside:
Closes #244
Closes #233
Closes #229 

The idea with the change is to have the step creating the PiP window in parallel:

Let Picture-in-Picture window be a new instance of PictureInPictureWindow associated with pictureInPictureElement.

And only after that, queueing the global task that does DOM mutations and what not. Because this creation of the Picture-in-picture can theoretically take forever.

This addresses the issues (and the WIP laid out there) in #244 which points out in-parallel mutation of the DOM and resolving promises when it can't.

I think this model fits what Chrome does already and being the one implementing this in Firefox, I feel comfortable with claiming it's achievable in some form there too.

cc @zcorpan && @beaufortfrancois - leaving PR to start discussion.

***
<a href="https://pr-preview.s3.amazonaws.com/theIDinside/picture-in-picture/pull/245.html" title="Last updated on May 10, 2026, 1:45 PM UTC (e4a9306)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/245/2983289...theIDinside:e4a9306.html" title="Last updated on May 10, 2026, 1:45 PM UTC (e4a9306)">Diff</a>

PR #260, opened 2026-06-16
Overhaul exit Picture-in-Picture algorithm

theIDinside:
Rework the exit Picture-in-Picture algorithm to mirror the requestPictureInPicture() overhaul (#245):
- run the work on the document's picture-in-picture parallel queue
- perform DOM mutation, promise resolution inside a queued global task
- keep the window close in parallel since closing may be asynchronous/disjoint across implementations.

The algorithm now takes |doc| and a promise-or-null |p| as input arguments.

First commit does not deal with:
- Fixing "unloading document steps". Unfortunately it can not re-use the algorithm, because it runs on the event loop, it needs to mutate DOM stuff and then enqueue work on the parallel queue to close the window.
- disablePictureInPicture must call the new algorithm by enqueing steps, providing input args
- when fullscreen flag is set, it must call the new algorithm by enqueing steps, providing input args

Addresses #252

***
<a href="https://pr-preview.s3.amazonaws.com/theIDinside/picture-in-picture/pull/260.html" title="Last updated on Jun 29, 2026, 5:08 PM UTC (81e4c56)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/picture-in-picture/260/9d88d7b...theIDinside:81e4c56.html" title="Last updated on Jun 29, 2026, 5:08 PM UTC (81e4c56)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -221,21 +221,32 @@ Picture-in-Picture window.
 
 ## Exit Picture-in-Picture ## {#exit-pip}
 
-When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked,
-the user agent MUST run the following steps:
+When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and promise-or-null
+|p|, run these steps:
 
-1. If {{pictureInPictureElement}} is `null`, throw a {{InvalidStateError}} and
-    abort these steps.
-2. Run the <a>close window algorithm</a> with the <a>Picture-in-Picture
-    window</a> associated with {{pictureInPictureElement}}.
-3. <a>Queue a task</a> to <a>fire an event</a> named
-    {{leavepictureinpicture}} using {{PictureInPictureEvent}} at the
-    |video| with its {{bubbles}} attribute initialized to `true` and its
-    {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
-    <a>Picture-in-Picture window</a> associated with {{pictureInPictureElement}}.
-4. Unset {{pictureInPictureElement}}.
-5. Remove one <a>item</a> matching <a>relevant settings object</a>'s <a>origin</a> from
-    <a>initiators of active Picture-in-Picture sessions</a>.
+1. Assert that |doc|'s [=Picture-in-Picture element=] is not `null`.
+2. Assert that these steps are running on the [=picture-in-picture parallel queue=].
+
+    Note: This algorithm is called from {{exitPictureInPicture}} and {{requestPictureInPicture}}.
+    The document unloading steps have to be structured differently due to running on the event loop
+    and can therefore not call this algorithm.
+3. Let |global| be |doc|'s [=relevant global object=].
+4. Run the <a>close window algorithm</a> with the <a>Picture-in-Picture window</a> associated with
+   |doc|'s [=Picture-in-Picture element=].
+5. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform the
+   following steps:
+    1. If |doc|'s [=Picture-in-Picture element=] is `null`:
+        1. [=/Resolve=] |p| if |p| is not `null`.
+        2. Return.
+    2. Let |element| be |doc|'s [=Picture-in-Picture element=].
+    3. Set |doc|'s [=Picture-in-Picture element=] to `null`.
+    4. <a>Fire an event</a> named {{leavepictureinpicture}} using {{PictureInPictureEvent}} at the
+       |element| with its {{bubbles}} attribute initialized to `true` and its
+       {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
+       <a>Picture-in-Picture window</a> associated with |element|.
+    5. Remove one <a>item</a> matching <a>relevant settings object</a>'s <a>origin</a> from
+       <a>initiators of active Picture-in-Picture sessions</a>.
+    6. [=/Resolve=] |p| if |p| is not `null`.
 
 It is NOT RECOMMENDED that the video playback state changes when the <a>exit
 Picture-in-Picture algorithm</a> is invoked. The website SHOULD be in control
@@ -340,18 +351,19 @@ The {{requestPictureInPicture()}} method steps <dfn export>request Picture-in-Pi
             [=/reject=] |p| with {{InvalidStateError}} {{DOMException}}.
         2. Abort these steps.
     4. Let |pipWindow| be a new instance of {{PictureInPictureWindow}} that represents [=this=]'s associated [=Picture-in-Picture window=].
-    5. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform
-        the following steps:
-        1. If {{pictureInPictureElement}} is not `null`, run the [=exit Picture-in-Picture algorithm=].
-        2. Set |doc|'s [=Picture-in-Picture element=] to [=this=].
-        3. [=list/Append=] [=relevant settings object=]'s [=origin=] to [=initiators of active
-            Picture-in-Picture sessions=].
-        4. If [=this=] is [=fullscreenElement=], [=exit fullscreen=].
-        5. [=Fire an event=] named {{enterpictureinpicture}} using {{PictureInPictureEvent}} at
-            [=this=] with its {{bubbles}} attribute initialized to `true` and its
-            {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
-            [=Picture-in-Picture window=].
-        6. [=/Resolve=] |p| with |pipWindow|.
+    5. If |doc|'s [=Picture-in-Picture element=] is not `null`, run the [=exit Picture-in-Picture
+       algorithm=] given |doc| and `null`.
+    6. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform
+       the following steps:
+        1. Set |doc|'s [=Picture-in-Picture element=] to [=this=].
+        2. [=list/Append=] [=relevant settings object=]'s [=origin=] to [=initiators of active
+           Picture-in-Picture sessions=].
+        3. If [=this=] is [=fullscreenElement=], [=exit fullscreen=].
+        4. [=Fire an event=] named {{enterpictureinpicture}} using {{PictureInPictureEvent}} at
+           [=this=] with its {{bubbles}} attribute initialized to `true` and its
+           {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
+           [=Picture-in-Picture window=].
+        5. [=/Resolve=] |p| with |pipWindow|.
 
 
 ## Extensions to <code>Document</code> ## {#document-extensions}
@@ -372,14 +384,19 @@ The {{pictureInPictureEnabled}} attribute's getter must return `true` if
 <dfn>Picture-in-Picture support</dfn> is `false` if there's a user preference
 that disables it or a platform limitation. It is `true` otherwise.
 
-The {{exitPictureInPicture()}} method, when invoked, MUST
-return <a>a new promise</a> |promise| and run the following steps <a>in
-parallel</a>:
+The {{exitPictureInPicture()}} method, when invoked, MUST return <a>a new promise</a> and run the
+following steps:
 
-1. Run the <a>exit Picture-in-Picture algorithm</a>.
-2. If the previous step threw an exception, reject |promise| with that
-    exception and abort these steps.
-3. [=/Resolve=] |promise|.
+1. If [=this=]'s [=Picture-in-Picture element=] is `null`, return [=a promise rejected with=]
+   {{InvalidStateError}} {{DOMException}}.
+2. Let |p| be [=a new promise=] created in [=this=]'s [=relevant realm=].
+3. Return |p|, and [=enqueue the following steps=] to [=this=]'s [=picture-in-picture parallel
+   queue=]:
+    1. If [=this=]'s [=Picture-in-Picture element=] is `null`:
+        1. [=Queue a global task=] on the [=media element event task source=] given [=this=]'s
+           [=relevant global object=] to [=/resolve=] |p|.
+        2. Return.
+    2. Run the <a>exit Picture-in-Picture algorithm</a> given [=this=] and |p|.
 
 ## Extension to <code>DocumentOrShadowRoot</code> ## {#documentorshadowroot-extension}
 
--- a/index.bs
+++ b/index.bs
@@ -221,8 +221,8 @@ Picture-in-Picture window.
 
 ## Exit Picture-in-Picture ## {#exit-pip}
 
-When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and nullable {{Promise}}
-|p|, run these steps:
+When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and nullable
+{{Promise}} |p|, run these steps:
 
 1. [=Assert=] that |doc|'s [=Picture-in-Picture element=] is not `null`.
 2. [=Assert=] that these steps are running on the [=picture-in-picture parallel queue=].
@@ -231,7 +231,7 @@ When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and
     |doc|'s [=Picture-in-Picture element=].
 5. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform the
     following steps:
-    1. If |doc|'s [=Picture-in-Picture element=] is `null`:
+    1. If |doc| is not [=fully active=] or |doc|'s [=Picture-in-Picture element=] is `null`:
         1. If |p| is not `null`, [=/resolve=] |p| with undefined.
         2. Return.
     2. Let |element| be |doc|'s [=Picture-in-Picture element=].
@@ -278,8 +278,12 @@ the user agent MAY run these steps:
 
 ## Interaction with Fullscreen ## {#fullscreen}
 
-It is RECOMMENDED to run the <a>exit Picture-in-Picture algorithm</a> when the
-{{pictureInPictureElement}} <a>fullscreen flag</a> is set.
+It is RECOMMENDED that when a [=Picture-in-Picture element=]'s <a>fullscreen flag</a> is set, the
+user agent runs these steps:
+
+1. Let |doc| be the [=Picture-in-Picture element=]'s [=node document=].
+2. [=Enqueue the following steps=] to |doc|'s <a>picture-in-picture parallel queue</a>:
+    1. Run the <a>exit Picture-in-Picture algorithm</a> given |doc| and `null`.
 
 ## Interaction with Page Visibility ## {#page-visibility}
 
@@ -382,8 +386,8 @@ that disables it or a platform limitation. It is `true` otherwise.
 
 The {{exitPictureInPicture()}} method, when invoked, runs the following steps:
 
-1. If [=this=]'s [=Picture-in-Picture element=] is `null`, return [=a promise rejected with=]
-    {{InvalidStateError}} {{DOMException}}.
+1. If [=this=]'s [=Picture-in-Picture element=] is `null` or [=this=] is not [=fully active=],
+    return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
 2. Let |p| be [=a new promise=] created in [=this=]'s [=relevant realm=].
 3. Return |p|, and [=enqueue the following steps=] to [=this=]'s [=picture-in-picture parallel
     queue=]:
--- a/index.bs
+++ b/index.bs
@@ -236,7 +236,7 @@ When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and
 5. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform the
     following steps:
     1. If |doc|'s [=Picture-in-Picture element=] is `null`:
-        1. [=/Resolve=] |p| if |p| is not `null`.
+        1. If |p| is not `null`, [=/resolve=] |p| with undefined.
         2. Return.
     2. Let |element| be |doc|'s [=Picture-in-Picture element=].
     3. Set |doc|'s [=Picture-in-Picture element=] to `null`.
--- a/index.bs
+++ b/index.bs
@@ -246,7 +246,7 @@ When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and
         <a>Picture-in-Picture window</a> associated with |element|.
     5. Remove one <a>item</a> matching <a>relevant settings object</a>'s <a>origin</a> from
         <a>initiators of active Picture-in-Picture sessions</a>.
-    6. [=/Resolve=] |p| if |p| is not `null`.
+    6. If |p| is not `null`, [=/resolve=] |p| with undefined.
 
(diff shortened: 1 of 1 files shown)


theIDinside:
Hmm can the spec not be formatted using `specfmt`?

theIDinside:
@chrisn mentioned in #252 that we should update the "unloading steps" to call the exit algorithm, but unfortunately this algorithm turned out to be trickier to actually formulate than I had anticipated (though I think it's in a good/better state now).

I don't know if it's possible to actually call it from the unloading steps, because we also need to run the "window closing" steps, which just like for `requestPictureInPicture` is one of those thing that have to run on the queue and take forever, so I was thinking for the unloading steps, we would have to just change the "normal" order (the order when user says `doc.exitPictureInPicture()` or when being called from with `el2.requestPictureInPicture()`), from
1. "close window"
2. "Queue task to mutate DOM stuff"
To the reverse, 
1. "mutate DOM stuff" (we are on the event loop so probably no task needed)
2. Enqueue steps to run on the parallel queue to close the window. (this can take forever. Needs to run on the queue)

I can add this change in a follow-up commit in this PR if that's preferable from a reviewer perspective

cc: @chrisn || @marcoscaceres

index.bs
```diff
@@ -222,21 +221,32 @@ Picture-in-Picture window.
...
-    <a>initiators of active Picture-in-Picture sessions</a>.
+When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and promise-or-null
+|p|, run these steps:
+
+1. Assert that |doc|'s [=Picture-in-Picture element=] is not `null`.
+2. Assert that these steps are running on the [=picture-in-picture parallel queue=].
+
+    Note: This algorithm is called from {{exitPictureInPicture}} and {{requestPictureInPicture}}.
+    The document unloading steps have to be structured differently due to running on the event loop
+    and can therefore not call this algorithm.
```
theIDinside: This is the comment I left in the main thread with respect to unloading steps. I think maybe this `Note` doesn't really belong in the spec and should probably go on second thought.

index.bs
```diff
@@ -222,21 +221,32 @@ Picture-in-Picture window.
...
-    window</a> associated with {{pictureInPictureElement}}.
-3. <a>Queue a task</a> to <a>fire an event</a> named
-    {{leavepictureinpicture}} using {{PictureInPictureEvent}} at the
-    |video| with its {{bubbles}} attribute initialized to `true` and its
-    {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
-    <a>Picture-in-Picture window</a> associated with {{pictureInPictureElement}}.
-4. Unset {{pictureInPictureElement}}.
-5. Remove one <a>item</a> matching <a>relevant settings object</a>'s <a>origin</a> from
-    <a>initiators of active Picture-in-Picture sessions</a>.
+When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and promise-or-null
```
marcoscaceres: Maybe: 
```suggestion
When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc| and nullable {{Promise}}
```

index.bs
```diff
@@ -222,21 +221,32 @@ Picture-in-Picture window.
...
+|p|, run these steps:
+
+1. Assert that |doc|'s [=Picture-in-Picture element=] is not `null`.
+2. Assert that these steps are running on the [=picture-in-picture parallel queue=].
```
```suggestion
1. [=Assert=] that |doc|'s [=Picture-in-Picture element=] is not `null`.
2. [=Assert=] that these steps are running on the [=picture-in-picture parallel queue=].
```

index.bs
marcoscaceres: Yeah, fine to get rid of it.

index.bs
```diff
@@ -222,21 +221,32 @@ Picture-in-Picture window.
...
+    Note: This algorithm is called from {{exitPictureInPicture}} and {{requestPictureInPicture}}.
+    The document unloading steps have to be structured differently due to running on the event loop
+    and can therefore not call this algorithm.
+3. Let |global| be |doc|'s [=relevant global object=].
+4. Run the <a>close window algorithm</a> with the <a>Picture-in-Picture window</a> associated with
+    |doc|'s [=Picture-in-Picture element=].
+5. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform the
+    following steps:
+    1. If |doc|'s [=Picture-in-Picture element=] is `null`:
+        1. [=/Resolve=] |p| if |p| is not `null`.
```
```suggestion
        1. If |p| is not `null`, [=/resolve=] |p| with undefined.
```

index.bs
```diff
@@ -222,21 +221,32 @@ Picture-in-Picture window.
...
+        2. Return.
+    2. Let |element| be |doc|'s [=Picture-in-Picture element=].
```
marcoscaceres, line 236: Can element be null?

index.bs
```diff
@@ -222,21 +221,32 @@ Picture-in-Picture window.
...
+    3. Set |doc|'s [=Picture-in-Picture element=] to `null`.
+    4. <a>Fire an event</a> named {{leavepictureinpicture}} using {{PictureInPictureEvent}} at the
+        |element| with its {{bubbles}} attribute initialized to `true` and its
+        {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
+        <a>Picture-in-Picture window</a> associated with |element|.
+    5. Remove one <a>item</a> matching <a>relevant settings object</a>'s <a>origin</a> from
+        <a>initiators of active Picture-in-Picture sessions</a>.
+    6. [=/Resolve=] |p| if |p| is not `null`.
```
```suggestion
    6. If |p| is not `null`, [=/resolve=] |p| with undefined.
```

index.bs
```diff
@@ -341,18 +351,19 @@ The {{requestPictureInPicture()}} method steps <dfn export>request Picture-in-Pi
...
         the following steps:
-        1. If {{pictureInPictureElement}} is not `null`, run the [=exit Picture-in-Picture algorithm=].
-        2. Set |doc|'s [=Picture-in-Picture element=] to [=this=].
-        3. [=list/Append=] [=relevant settings object=]'s [=origin=] to [=initiators of active
+        1. Set |doc|'s [=Picture-in-Picture element=] to [=this=].
+        2. [=list/Append=] [=relevant settings object=]'s [=origin=] to [=initiators of active
             Picture-in-Picture sessions=].
-        4. If [=this=] is [=fullscreenElement=], [=exit fullscreen=].
-        5. [=Fire an event=] named {{enterpictureinpicture}} using {{PictureInPictureEvent}} at
+        3. If [=this=] is [=fullscreenElement=], [=exit fullscreen=].
```
```suggestion
        3. If [=this=] is [=fullscreenElement=], then [=exit fullscreen=].
```

index.bs
```diff
@@ -373,14 +384,19 @@ The {{pictureInPictureEnabled}} attribute's getter must return `true` if
...
-The {{exitPictureInPicture()}} method, when invoked, MUST
-return <a>a new promise</a> |promise| and run the following steps <a>in
-parallel</a>:
-
-1. Run the <a>exit Picture-in-Picture algorithm</a>.
-2. If the previous step threw an exception, reject |promise| with that
-    exception and abort these steps.
-3. [=/Resolve=] |promise|.
+The {{exitPictureInPicture()}} method, when invoked, MUST return <a>a new promise</a> and run the
+following steps:
```
```suggestion
The {{exitPictureInPicture()}} method, when invoked, runs the following steps:
```

marcoscaceres:
The question of unloading might be one better handled through the document becoming non-fully active... if the document becomes not fully active, exit full screen is the observable behavior on Chrome and WebKit (PiP closes). 

Could we use that that instead of the unloading steps? @chrisn and I tested this using a little demo page. https://marcoscaceres.github.io/playground/pip-test/

chrisn:
Web Platform Design Principles has some detail on non-fully active: https://www.w3.org/TR/design-principles/#handle-non-fully-active-documents. But I'm not sure where we can hook into on the transition from fully active to non-fully active.

chrisn:
> I don't know if it's possible to actually call it from the unloading steps, because we also need to run the "window closing" steps, which just like for `requestPictureInPicture` is one of those thing that have to run on the queue and take forever, so I was thinking for the unloading steps, we would have to just change the "normal" order (the order when user says `doc.exitPictureInPicture()` or when being called from with `el2.requestPictureInPicture()`), from
> 
> 1. "close window"
> 2. "Queue task to mutate DOM stuff"
>    To the reverse,
> 3. "mutate DOM stuff" (we are on the event loop so probably no task needed)
> 4. Enqueue steps to run on the parallel queue to close the window. (this can take forever. Needs to run on the queue)
> 
> I can add this change in a follow-up commit in this PR if that's preferable from a reviewer perspective

Makes sense to me, and adding to this PR sounds fine.

marcoscaceres:
Following up on @chrisn's point about handling this through the document becoming non-fully active, and the [TAG Design Principles §2.13](https://www.w3.org/TR/design-principles/#handle-non-fully-active-documents) he linked.

I think this PR (and the spec as it stands) has a gap worth closing as part of the exit overhaul: nothing defines what happens to an active Picture-in-Picture session when its document stops being [fully active](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active), that is, when the containing `<iframe>` is removed or navigated, on top-level navigation, or when the document enters bfcache. The reworked algorithm only guards on `pictureInPictureElement` being null; it never references "fully active". Deferring the unloading-steps integration from #252 is fine, but I'd like us to agree on the model now so the deferred work has a target.

On @chrisn's open question (where can we hook into the transition from fully active to non-fully active?): there isn't a single event for it, but Fullscreen already solves the same problem and is a good precedent. It hooks the two concrete algorithms that cause non-fully-active, namely the DOM [removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext) (element or subtree removed) and the HTML [unloading document cleanup steps](https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-document-cleanup-steps) (document unloaded or navigated), and it also fails fast at call time if the document isn't fully active.

Mapping §2.13 onto exit PiP, I'd suggest we define three things:

1. At call time: if the relevant document is not fully active, `exitPictureInPicture()` (and the exit-on-replace path inside `requestPictureInPicture()`) should fail immediately. §2.13 suggests `InvalidStateError`. Today we only check `pictureInPictureElement`.
2. When the queued global task runs: an explicit "if the document is not fully active, stop with no observable effects" check, rather than leaning on the `pictureInPictureElement is null` proxy (§2.13 also says not to rely on garbage collection for this).
3. On the transition itself: define the teardown by hooking the removing steps and the unloading document cleanup steps (the Fullscreen pattern), and spell out what happens to `pictureInPictureElement`, whether `leavepictureinpicture` fires, and how any in-flight promise settles. That is all currently left to the implementation.

One thing for the group on (3): when a document becomes non-fully active mid-exit, should the returned promise reject with `AbortError` (§2.13's general rule for aborted async operations), or resolve, since the requested end state ("no longer in PiP") is arguably already satisfied by the teardown? I lean toward making it definite and observable either way; I don't think the current text forces a single outcome.

I put together a small test bench for these transitions (remove iframe, navigate iframe, top-level navigation, bfcache restore, and a parent holding a `PictureInPictureWindow` across destruction): https://marcoscaceres.github.io/playground/pip-test/. In my testing the behaviour across these paths isn't consistent today, which is the main reason I'd rather we define the model than leave it implementation-defined. It's set up so you can check your own engine.

marcoscaceres:
WebKit tracking bug https://bugs.webkit.org/show_bug.cgi?id=317409

theIDinside:
> The question of unloading might be one better handled through the document becoming non-fully active... if the document becomes not fully active, exit full screen is the observable behavior on Chrome and WebKit (PiP closes).
> 
> Could we use that that instead of the unloading steps? @chrisn and I tested this using a little demo page. https://marcoscaceres.github.io/playground/pip-test/

This is fantastic. Could you maybe add an example from https://github.com/w3c/picture-in-picture/issues/261 - basically a button that both does getDisplayMedia  + requestPip. Effectively, it may end up being the case that, that *should* fail (though it doesn't on safari and chrome), given that these two requests are spec'ed on consuming and being gated by user activation.

theIDinside:
> Following up on @chrisn's point about handling this through the document becoming non-fully active, and the [TAG Design Principles §2.13](https://www.w3.org/TR/design-principles/#handle-non-fully-active-documents) he linked.
> 
> I think this PR (and the spec as it stands) has a gap worth closing as part of the exit overhaul: nothing defines what happens to an active Picture-in-Picture session when its document stops being [fully active](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active), that is, when the containing `<iframe>` is removed or navigated, on top-level navigation, or when the document enters bfcache. 
> ...
> On @chrisn's open question (where can we hook into the transition from fully active to non-fully active?): there isn't a single event for it, but Fullscreen already solves the same problem and is a good precedent. It hooks the two concrete algorithms that cause non-fully-active, namely the DOM [removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext) (element or subtree removed) and the HTML [unloading document cleanup steps](https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-document-cleanup-steps) (document unloaded or navigated), and it also fails fast at call time if the document isn't fully active.

So can we really hook into DOM [removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext) (element or subtree removed)? We have tests historically that rely on unbound elements (and also being able to remove the actual element being PIP'ed). And in the case of an iframe; would the removal of the iframe eventually also mean that the HTML [unloading document cleanup steps](https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-document-cleanup-steps) run? (thus having our "not fully active"-case be caught fully by the unloading steps?)

> Mapping §2.13 onto exit PiP, I'd suggest we define three things:
> 
>     1. At call time: if the relevant document is not fully active, `exitPictureInPicture()` (and the exit-on-replace path inside `requestPictureInPicture()`) should fail immediately. §2.13 suggests `InvalidStateError`. Today we only check `pictureInPictureElement`.
> 
>     2. When the queued global task runs: an explicit "if the document is not fully active, stop with no observable effects" check, rather than leaning on the `pictureInPictureElement is null` proxy (§2.13 also says not to rely on garbage collection for this).

Adding these checks to exit and has filed an issue for the requestPIP part. And let's follow the recommendations; `InvalidStateError` for sync/pre-checks. Currently I've made the async-part resolve with undefined as per the tear down of state being satisfied, even if being different from recommendations I think we can argue that striving for simplicity also is a goal. But no hard requirement on my part.

theIDinside:
Added the unloading steps in the last commit. I think that at least partially addresses most of your concerns @marcoscaceres ?

index.bs
```diff
@@ -373,14 +391,18 @@ The {{pictureInPictureEnabled}} attribute's getter must return `true` if
...
-parallel</a>:
-
-1. Run the <a>exit Picture-in-Picture algorithm</a>.
-2. If the previous step threw an exception, reject |promise| with that
-    exception and abort these steps.
-3. [=/Resolve=] |promise|.
+The {{exitPictureInPicture()}} method, when invoked, runs the following steps:
+
+1. If [=this=]'s [=Picture-in-Picture element=] is `null` or [=this=] is not [=fully active=],
+    return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
```
beaufortfrancois, line 397: Nit:

```suggestion
1. If [=this=] is not [=fully active=],
    return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.
2. If [=this=]'s [=Picture-in-Picture element=] is `null`,
    return [=a promise rejected with=] {{InvalidStateError}} {{DOMException}}.

Comment (beaufortfrancois) [reply index.bs:397]: Can we also have web-platform tests along this PR to help figuring out what needs to be fixed in different implementations?

File: index.bs [hunk 10]
```diff
@@ -222,30 +221,44 @@ Picture-in-Picture window.
...
+{{Promise}} |p|, run these steps:
+
+1. [=Assert=] that these steps are running on the [=picture-in-picture parallel queue=].
+2. Let |global| be |doc|'s [=relevant global object=].
+3. Run the <a>close window algorithm</a> with the <a>Picture-in-Picture window</a> associated with
+    |doc|'s [=Picture-in-Picture element=].
+4. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform the
+    following steps:
+    1. If |doc| is not [=fully active=] or |doc|'s [=Picture-in-Picture element=] is `null`:
+        1. If |p| is not `null`, [=/resolve=] |p| with undefined.
```
beaufortfrancois, line 234: I naively thought we should reject instead of resolve promise. Why not in this case?

beaufortfrancois (changes requested):

index.bs
```diff
@@ -222,30 +221,44 @@ Picture-in-Picture window.
...
-    window</a> associated with {{pictureInPictureElement}}.
-3. <a>Queue a task</a> to <a>fire an event</a> named
-    {{leavepictureinpicture}} using {{PictureInPictureEvent}} at the
-    |video| with its {{bubbles}} attribute initialized to `true` and its
-    {{PictureInPictureEvent/pictureInPictureWindow}} attribute initialized to
-    <a>Picture-in-Picture window</a> associated with {{pictureInPictureElement}}.
-4. Unset {{pictureInPictureElement}}.
-5. Remove one <a>item</a> matching <a>relevant settings object</a>'s <a>origin</a> from
-    <a>initiators of active Picture-in-Picture sessions</a>.
+When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given |doc|,  and nullable
```
marcoscaceres, line 224:
```suggestion
When the <dfn>exit Picture-in-Picture algorithm</dfn> is invoked given a {{Document}} |doc|,  and nullable
```

index.bs
marcoscaceres: Agree... this should reject with an InvalidStateError.

marcoscaceres: Also, these should happen both before [=Queue a global task=] and recheck after [=Queue a global task=]. 

See https://www.w3.org/TR/design-principles/#support-non-fully-active

index.bs
```diff
@@ -222,30 +221,44 @@ Picture-in-Picture window.
...
+3. Run the <a>close window algorithm</a> with the <a>Picture-in-Picture window</a> associated with
+    |doc|'s [=Picture-in-Picture element=].
+4. [=Queue a global task=] on the [=media element event task source=] given |global|, to perform the
+    following steps:
+    1. If |doc| is not [=fully active=] or |doc|'s [=Picture-in-Picture element=] is `null`:
+        1. If |p| is not `null`, [=/resolve=] |p| with undefined.
+        2. Return.
+    2. Let |element| be |doc|'s [=Picture-in-Picture element=].
+    3. Set |doc|'s [=Picture-in-Picture element=] to `null`.
+    4. <a>Fire an event</a> named {{leavepictureinpicture}} using {{PictureInPictureEvent}} at the
```
marcoscaceres, line 238: [=Queue a task=] to perform the follow steps?

index.bs
```diff
@@ -222,30 +221,44 @@ Picture-in-Picture window.
...
 It is NOT RECOMMENDED that the video playback state changes when the <a>exit
 Picture-in-Picture algorithm</a> is invoked. The website SHOULD be in control
 of the experience if it is website initiated. However, the user agent MAY expose
 Picture-in-Picture window controls that change video playback state (e.g.,
 pause).
 
-As one of the <a>unloading document cleanup steps</a>, run the <a>exit
-Picture-in-Picture algorithm</a>.
+### Unloading steps ### {#unloading-steps}
+As one of the <a>unloading document cleanup steps</a> given |doc| run the following steps:
```
marcoscaceres, line 253:
```suggestion
As one of the <a>unloading document cleanup steps</a> given {{Document}} |doc| run the following steps:
```

index.bs
```diff
@@ -272,8 +285,12 @@ the user agent MAY run these steps:
 
 ## Interaction with Fullscreen ## {#fullscreen}
 
-It is RECOMMENDED to run the <a>exit Picture-in-Picture algorithm</a> when the
-{{pictureInPictureElement}} <a>fullscreen flag</a> is set.
+It is RECOMMENDED that when a [=Picture-in-Picture element=]'s <a>fullscreen flag</a> is set, the
```
marcoscaceres, line 288: I think we talked about this, so apologies... why is the RECOMMENDED instead of MUST? 

Can we add a note here explaining why it's not a MUST requirement?

marcoscaceres (changes requested):

theIDinside:
So I just realized something @marcoscaceres | @chrisn | @beaufortfrancois  while working on other stuff in gecko, I was reminded of the idea of [runnable task](https://html.spec.whatwg.org/multipage/webappapis.html#concept-task-runnable). We (Firefox) had a bug in `postMessage`, which posted a message from inside an `unload` handler, which would be seen by a "new" document upon reload. We don't have the concept of queuing a task like blink or webkit does as you may know, but effectively doing such a check was the fix (if document is not fully active, don't run the work posted to our internal implementation of dispatched work)

We queue tasks to run from the PIP parallel queue. And per spec [these should not run if the document is not fully active](https://html.spec.whatwg.org/multipage/webappapis.html#concept-task-runnable:~:text=If%20the%20event%20loop%20has%20a%20task%20queue%20with%20at%20least%20one%20runnable%20task%3A) which means that our "is fully active" is entirely superfluous. Because the task should not be started if the document itself is no longer active.

index.bs
theIDinside: Left a comment in the main thread, but I think there's an argument for that we should not be doing this check at all.

index.bs
theIDinside: I actually wrote a postMessage web platform tests *specifically* for this "not fully active" scenario, and it involves actually testing that something "did not happen" - and we wouldn't be able to test this here,  I think. This promise will never get resolve or rejected, I believe. It should *just* be doing the null check.

index.bs
chrisn, line 288: Need to invoke this from step 3 in https://fullscreen.spec.whatwg.org/#fullscreen-an-element

chrisn: See also https://github.com/w3c/picture-in-picture/issues/191. But let's work on that in a follow up PR.

marcoscaceres: Right, yeah... it's basically transitioning in/out of fullscreen otherwise PiP,

index.bs
```diff
@@ -222,30 +221,44 @@ Picture-in-Picture window.
...
 
 It is NOT RECOMMENDED that the video playback state changes when the <a>exit
 Picture-in-Picture algorithm</a> is invoked. The website SHOULD be in control
 of the experience if it is website initiated. However, the user agent MAY expose
 Picture-in-Picture window controls that change video playback state (e.g.,
 pause).
 
-As one of the <a>unloading document cleanup steps</a>, run the <a>exit
-Picture-in-Picture algorithm</a>.
+### Unloading steps ### {#unloading-steps}
```
marcoscaceres, line 252: Talking to @chrisn, would be great to just move these over to HTML.

index.bs
chrisn, line 253:
```suggestion
The <dfn export>picture in picture unloading steps</dfn>, given {{Document}} |doc| are:
```
Exported so we can call this from HTML

marcoscaceres:
@theIDinside added what you requested above to https://marcoscaceres.github.io/playground/pip-test/

index.bs
chrisn: See https://github.com/whatwg/html/pull/12689 for the corresponding HTML change.

